PR #29, opened 2022-05-23
Initial ViT Implementation & Pretrained ImageNet Models

avik-pal:
cc @vchuravy


lib/Boltz/src/utils.jl
```diff
@@ -0,0 +1,26 @@
+"""
+    fast_chunk(x::AbstractArray, ::Val{n}, ::Val{dim})
+
+Type-stable and faster version of `MLUtils.chunk`
+"""
+@inline fast_chunk(h::Int, n::Int) = (1:h) .+ h * (n - 1)
+@inline fast_chunk(x::AbstractArray, h::Int, n::Int, ::Val{dim}) where {dim} = selectdim(x, dim, fast_chunk(h, n))
+@inline fast_chunk(x::AbstractArray, ::Val{N}, d::Val{D}) where {N,D} = fast_chunk.((x,), size(x, D) ÷ N, 1:N, d)
```
ToucheSir, line 8: Any reason this can't be upstreamed? :)

avik-pal: Yeah we can do it. It will be a breaking API change though to use `Val` instead of integers

PR #40, opened 2022-05-28
Extending `Scale` to allow for multiple dimension inputs

theabhirath:
This PR extends `Scale` to work with multiple inputs (this might be useful for `LayerNorm` when implemented, and also achieves feature parity with Flux in this regard). It also adds tests for `Scale`. This is my first contribution here, so any feedback if I've done something wrong (given that I'm more familiar with the Flux codebase) will be appreciated 😅


avik-pal:
Thanks for the contribution. Can you make 2 additional updates:
1. Bump the patch version in Project.toml
2. Update the docs for `Scale`. Also add a note that multiple dims requires a version > (new version)

PR #56, opened 2022-06-18
Cleaner and Better Documentation

avik-pal:
TODOs:
- [x] Use `@doc doc` for docstrings containing LaTeX
- [x] Shift to MKDocs Material theme (it is just prettier)
  - [ ] Can we make the functions more prominently highlighted?
  - [x] Code highlighting?
- [x] Restructure Parts of the Documentation
  - [x] Users don't know what to do to integrate Lux with their framework (See https://github.com/SciML/DiffEqFlux.jl/issues/727). Make the AbstractLayer API more prominent
  - [x] Update Style Guide to document deviations from SciMLStyle
- [x] Add docs for packages in `lib/` 
- [x] Turn off draft mode before merging


src/core.jl
```diff
@@ -6,6 +6,28 @@
...
+"""
+    AbstractExplicitLayer
+
+Abstract Type for all Lux Layers
+
+Users implementing their custom layer, **must** implement
+
+  - `initialparameters(rng::AbstractRNG, layer::CustomAbstractExplicitLayer)` -- This
+    returns a `NamedTuple` containing the trainable parameters for the layer.
+  - `initialstates(rng::AbstractRNG, layer::CustomAbstractExplicitLayer)` -- This returns a
```
MilkshakeForReal, line 18: My understanding is that if my custom layer does not have a state then I don't need to implement this function? Since it is calling `initialstates(::AbstractRNG, ::Any) = NamedTuple()`

PR #69, opened 2022-06-20
Manual to translate Flux to Lux

avik-pal:


avik-pal:
@ToucheSir what is the correct way of getting gradients for only trainable parameters in Flux? I know the params way http://lux.csail.mit.edu/previews/PR69/manual/migrate_from_flux/#implementing-custom-layers but want to avoid it.

docs/src/manual/migrate_from_flux.md
```diff
@@ -0,0 +1,213 @@
...
+    end
+
+
+
+
+
+
+
+    # `A` is not trainable
+    Flux.trainable(f::FluxLinear) = (f.B,)
```
ToucheSir, line 105: What you wrote is the way to go (default to `@functor [Layer] [fields...]` unless you have non-trainable state, in which case `trainable` is necessary as well (BTW this is exported by Optimisers now, so you can define layers without importing Flux at all. Perhaps that should be the recommendation).

The only change to make is this:
```suggestion
    Flux.trainable(f::FluxLinear) = (; B=f.B)
```
Not including prop names in `trainable` is deprecated and will be removed in the next major version of Optimisers.

ToucheSir: (1 hour during a work day lunch break is a little short notice for reviewing a PR 😂 )

avik-pal: I needed the manual to go through for a corresponding DiffEqSensitivity PR. I will update the example in a new PR.

PR #77, opened 2022-06-27
Fix ambiguous update!

IlyaOrson:
Fix https://github.com/FluxML/Optimisers.jl/issues/91


avik-pal:
@IlyaOrson can you check if https://github.com/avik-pal/Lux.jl/pull/74 fixes the problem? I removed the piracy all together

Issue #83, opened 2022-07-08
Lighter syntax for stateless networks?


gdalle:
When the neural network I'm working with has no evolving state, I find it quite cumbersome to always call `model(x, ps, st)`. This becomes especially painful when the `st` needs to be passed down as an argument, even though it is essentially useless.
How hard would it be to have a default `model(x, ps)` for these cases? Or a very cheap function `empty_state(model)` to avoid passing down `st` in lower level calls?

avik-pal:
There are a few solutions to this (believe me I want it too. It will be a significant QoL feature). But there are quite a few shortcomings of each:

1. https://github.com/avik-pal/Lux.jl/pull/46 -- Use code transformations so that users can define models without state and we automatically generate the required code. This is not very trivial to implement and there are still edge cases.
2. How about `(l::AbstractLayer)(x, ps, st) = l(x, ps), NamedTuple()`? In this case, the user might have defined `l(x, st)` but since we don't know parameters from states by their types this call is inherently ambiguous. Could we deal with this using better documentation? Possibly, but I have tried extremely hard not to make implicit assumptions about user code (except 3-4 interface requirements), but this would violate that.
3. Update abstract types to contain information like `AbstractExplicitLayer{paramterless, stateless}`. Then we can unambiguously implement point 2. In the long term, I see this as the most stable solution to the problem and will honestly simplify a lot of code, but it is also not on my current priority list to implement. (That being said I am happy to review and merge a PR implementing this)

gdalle:
I agree that option 3 seems like the most natural and achievable! Not sure I can submit a PR about this right now, but I'll still take a look

gdalle:
Just to be sure, the code to modify would only be `core.jl` and the `layers` folder?

PR #46, opened 2022-06-07
Add Lux.@layerdef for easy custom model creation

avik-pal:
## TODOs

- [ ] Highlight what this API cannot do yet
- [ ] Allow more configuration of the models by parsing the keyword arguments
- [x] Add a `contrib` page to documentation
- [ ] Documentation for the Breaking Change
  - [ ]  `update_guide.md`
  - [x] `ecosystem.md`
  - [x] `overview.md`
- [ ] Tests for `@compact`
- [x] Add a syntax to handle shared layers

## Lux.@compact

An Experimental compact API (inspired by `flax.linen.compact` decorator) to make it easier to write Lux models.

### Example

```julia
using Lux, Random, Zygote, NNlib

Lux.@compact function MyCustomModel((x, y)::T) where {T}
    @assert size(x, 1) == 2 && size(y, 1) == 2
    a = Dense(3, 3, relu)(Dense(2, 3)(x) .+ Dense(2, 3, relu)(x)) .- Dense(2, 3, gelu)(x)
    b = Dense(3, 2, sigmoid)(a)
    c = Dense(2, 2, relu)(y)
    return b .+ c
end

model = MyCustomModel()

ps, st = Lux.setup(Random.default_rng(), model)
x, y = randn(Float32, 2, 1), randn(Float32, 2, 1)

model((x, y), ps, st)
```

### Generated Model

```julia
quote
    struct MyCustomModel{__dense_1_type, __dense_2_type, __dense_3_type, __dense_4_type, __dense_5_type, __dense_6_type} <: Lux.AbstractExplicitContainerLayer{(:dense_1, :dense_2, :dense_3, :dense_4, :dense_5, :dense_6)}
        dense_1::__dense_1_type
        dense_2::__dense_2_type
        dense_3::__dense_3_type
        dense_4::__dense_4_type
        dense_5::__dense_5_type
        dense_6::__dense_6_type
    end
    function MyCustomModel(; )
        return MyCustomModel(Dense(3, 3, relu), Dense(2, 3), Dense(2, 3, relu), Dense(2, 3, gelu), Dense(3, 2, sigmoid), Dense(2, 2, relu))
    end
    function (mandrill::MyCustomModel)((x, y)::T, ps::Union{Lux.ComponentArray, NamedTuple}, st::NamedTuple; ) where T
        if size(x, 1) == 2 && size(y, 1) == 2
            nothing
        else
            Base.throw(Base.AssertionError("size(x, 1) == 2 && size(y, 1) == 2"))
        end
[...]

PR #84, opened 2022-07-09
Support layers without parameters or state

gdalle:
This PR proposes a solution to #83 by adding 2 booleans `P` and `S` to the layer types, specifying whether or not they need parameters / states as inputs.

That way, in addition to the generic syntax `layer(x, ps, st)`:
- a layer without parameters (e.g. `Dropout`) supports `layer(x, st)`
- a layer without state (e.g. `Dense`) supports `layer(x, ps)`
- a layer without parameters or state (e.g. `ReshapeLayer`) supports `layer(x)`

### DIFF
--- a/src/core.jl
+++ b/src/core.jl
@@ -101,15 +101,27 @@ function apply(model::AbstractExplicitLayer, x, ps, st::NamedTuple)
 end
 
 function apply(model::AbstractExplicitLayer{true, false}, x, ps)
-    return model(x, ps, NamedTuple())
+    return apply(model, x, ps, NamedTuple())
 end
 
 function apply(model::AbstractExplicitLayer{false, true}, x, st::NamedTuple)
-    return model(x, NamedTuple(), st)
+    return apply(model, x, NamedTuple(), st)
 end
 
 function apply(model::AbstractExplicitLayer{false, false}, x)
-    return model(x, NamedTuple(), NamedTuple())
+    return apply(model, x, NamedTuple(), NamedTuple())
+end
+
+function (model::AbstractExplicitLayer{true, false})(x, ps)
+    return apply(model, x, ps)
+end
+
+function (model::AbstractExplicitLayer{false, true})(x, st::NamedTuple)
+    return apply(model, x, st)
+end
+
+function (model::AbstractExplicitLayer{false, false})(x)
+    return apply(model, x)
 end
 
 function Base.show(io::IO, x::AbstractExplicitLayer)
--- a/src/core.jl
+++ b/src/core.jl
@@ -1,7 +1,12 @@
 """
-    AbstractExplicitLayer
+    AbstractExplicitLayer{hasparams,hasstate}
 
 Abstract Type for all Lux Layers
+
+Type parameters:
+
+  - `hasparams::Bool`: indicates that the layer requires parameters
+  - `hasstate::Bool`: indicates that the layer requires a state
 
 Users implementing their custom layer, **must** implement
 
@@ -20,7 +25,7 @@ Optionally:
 
 See also [`AbstractExplicitContainerLayer`](@ref)
 """
-abstract type AbstractExplicitLayer end
+abstract type AbstractExplicitLayer{hasparams, hasstate} end
 
 """
     initialparameters(rng::AbstractRNG, l)
@@ -28,6 +33,10 @@ abstract type AbstractExplicitLayer end
 Generate the initial parameters of the layer `l`.
 """
 initialparameters(::AbstractRNG, ::AbstractExplicitLayer) = NamedTuple()
+function initialparameters(::AbstractRNG,
+                           ::AbstractExplicitLayer{false, hasstate}) where {hasstate}
+    return NamedTuple()
+end
 function initialparameters(rng::AbstractRNG, l::NamedTuple)
     return map(Base.Fix1(initialparameters, rng), l)
 end
@@ -38,6 +47,10 @@ end
 Generate the initial states of the layer `l`.
 """
 initialstates(::AbstractRNG, ::AbstractExplicitLayer) = NamedTuple()
+function initialstates(::AbstractRNG,
+                       ::AbstractExplicitLayer{hasparams, false}) where {hasparams}
+    return NamedTuple()
+end
 initialstates(rng::AbstractRNG, l::NamedTuple) = map(Base.Fix1(initialstates, rng), l)
 
 """
@@ -47,6 +60,9 @@ Return the total number of parameters of the layer `l`.
 """
 function parameterlength(l::AbstractExplicitLayer)
     return parameterlength(initialparameters(Random.default_rng(), l))
+end
+function parameterlength(::AbstractExplicitLayer{hasstate, false}) where {hasstate}
+    return 0
 end
 function parameterlength(nt::Union{NamedTuple, Tuple})
     return length(nt) == 0 ? 0 : sum(parameterlength, nt)
@@ -59,6 +75,7 @@ parameterlength(a::AbstractArray) = length(a)
 Return the total number of states of the layer `l`.
 """
 statelength(l::AbstractExplicitLayer) = statelength(initialstates(Random.default_rng(), l))
+statelength(l::AbstractExplicitLayer{false, hasparams}) where {hasparams} = 0
 statelength(nt::Union{NamedTuple, Tuple}) = length(nt) == 0 ? 0 : sum(statelength, nt)
 statelength(a::AbstractArray) = length(a)
 statelength(x::Union{Number, Symbol, Val}) = 1
@@ -83,6 +100,19 @@ function apply(model::AbstractExplicitLayer, x, ps, st::NamedTuple)
     return model(x, ps, st)
 end
 
+function apply(model::AbstractExplicitLayer{hasparams, false}, x, ps) where {hasparams}
+    return model(x, ps, NamedTuple())
+end
+
+function apply(model::AbstractExplicitLayer{false, hasstate}, x,
+               st::NamedTuple) where {hasstate}
+    return model(x, NamedTuple(), st)
+end
+
+function apply(model::AbstractExplicitLayer{false, false}, x)
+    return model(x, NamedTuple(), NamedTuple())
+end
+
 function Base.show(io::IO, x::AbstractExplicitLayer)
     __t = rsplit(string(get_typename(x)), "."; limit=2)
     T = length(__t) == 2 ? __t[2] : __t[1]
@@ -91,20 +121,30 @@ end
 
 # Abstract Container Layers
 """
-    AbstractExplicitContainerLayer{layers} <: AbstractExplicitLayer
+    AbstractExplicitContainerLayer{layers,hasparams,hasstate} <: AbstractExplicitLayer{hasparams,hasstate}
 
 Abstract Container Type for certain Lux Layers. `layers` is a tuple containing fieldnames
 for the layer, and constructs the parameters and states using those.
 
 Users implementing their custom layer can extend the same functions as in
+initialstates(::AbstractRNG, ::AbstractExplicitLayer) = NamedTuple()
 [`AbstractExplicitLayer`](@ref)
 """
-abstract type AbstractExplicitContainerLayer{layers} <: AbstractExplicitLayer end
+abstract type AbstractExplicitContainerLayer{layers, hasparams, hasstate} <:
+              AbstractExplicitLayer{hasparams, hasstate} end
 
 function initialparameters(rng::AbstractRNG,
                            l::AbstractExplicitContainerLayer{layers}) where {layers}
     length(layers) == 1 && return initialparameters(rng, getfield(l, layers[1]))
     return NamedTuple{layers}(initialparameters.(rng, getfield.((l,), layers)))
+end
+
+function initialparameters(::AbstractRNG,
+                           ::AbstractExplicitContainerLayer{layers, false, hasstate}) where {
+                                                                                             layers,
+                                                                                             hasstate
+                                                                                             }
+    return NamedTuple()
 end
 
 function initialstates(rng::AbstractRNG,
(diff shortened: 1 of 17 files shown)


gdalle:
Sofar I haven't added specific tests for these new methods, and they are not documented either. This is just a first draft to agree on the syntax before moving forward

gdalle:
The main issue I see is for constructs such as `Parallel`, `BranchLayer`, `PairwiseFusion` and `Chain`. Since those layers are built from potentially heterogeneous `NamedTuple`s, I don't know how to statically determine whether they have parameters / states. One would need to apply an OR to the `hasparams` / `hasstate` of each element of the `NamedTuple`, which sounds like it has to be done dynamically since we don't even know its length.

I was thinking, would be recursive pairwise construction work? If a chain is stored as `Chain(Chain(l1, l2), l3)` instead of `Chain(l1, l2, l3)`, does that make it easier to propagate parameter / state information? Does it slow computations down?

gdalle:
> The main issue I see is for constructs such as `Parallel`, `BranchLayer`, `PairwiseFusion` and `Chain`.

I may have found a solution. A `Chain` (or any other similar construct) requires parameters / states if and only if at least one of its components does. If I'm right, the code below can decide that statically (just imagine `B` is either `hasparams` or `hasstate`):

```julia
julia> using Test

julia> struct MyLayer{B} end

julia> struct MyChain{B} end

julia> MyChain(t::Tuple{Vararg{MyLayer}}) = MyChainTuple{true}();

julia> MyChain(t::Tuple{Vararg{MyLayer{false}}}) = MyChainTuple{false}();

julia> l0, l1 = MyLayer{false}(), MyLayer{true}();

julia> c00 = @inferred MyChain((l0, l0))
MyChainTuple{false}()

julia> c01 = @inferred MyChain((l0, l1))
MyChainTuple{true}()

julia> c11 = @inferred MyChain((l1, l1))
MyChainTuple{true}()
```

src/core.jl
```diff
@@ -20,14 +25,18 @@ Optionally:
...
+abstract type AbstractExplicitLayer{hasparams, hasstate} end
 
 """
     initialparameters(rng::AbstractRNG, l)
 
 Generate the initial parameters of the layer `l`.
 """
 initialparameters(::AbstractRNG, ::AbstractExplicitLayer) = NamedTuple()
+function initialparameters(::AbstractRNG,
+                           ::AbstractExplicitLayer{false, hasstate}) where {hasstate}
```
avik-pal, line 37: Unused `hasstate`

src/core.jl
```diff
@@ -48,6 +61,9 @@ Return the total number of parameters of the layer `l`.
 function parameterlength(l::AbstractExplicitLayer)
     return parameterlength(initialparameters(Random.default_rng(), l))
 end
+function parameterlength(::AbstractExplicitLayer{hasstate, false}) where {hasstate}
```
avik-pal, line 64: Ordering is incorrect

src/core.jl
```diff
@@ -20,14 +25,18 @@ Optionally:
 
 See also [`AbstractExplicitContainerLayer`](@ref)
 """
-abstract type AbstractExplicitLayer end
+abstract type AbstractExplicitLayer{hasparams, hasstate} end
```
avik-pal, line 28: Let's add functions to check this as well. `hasstates(::AbstractExplicitLayer)` and `hasparams(::AbstractExplicitLayer)`

src/core.jl
```diff
@@ -59,6 +75,7 @@ parameterlength(a::AbstractArray) = length(a)
 Return the total number of states of the layer `l`.
 """
 statelength(l::AbstractExplicitLayer) = statelength(initialstates(Random.default_rng(), l))
+statelength(l::AbstractExplicitLayer{false, hasparams}) where {hasparams} = 0
```
avik-pal, line 78: Ordering

src/core.jl
```diff
@@ -83,6 +100,30 @@ function apply(model::AbstractExplicitLayer, x, ps, st::NamedTuple)
...
+function apply(model::AbstractExplicitLayer{true, false}, x, ps)
+    return apply(model, x, ps, NamedTuple())
+end
+
+function apply(model::AbstractExplicitLayer{false, true}, x, st::NamedTuple)
+    return apply(model, x, NamedTuple(), st)
+end
+
+function apply(model::AbstractExplicitLayer{false, false}, x)
+    return apply(model, x, NamedTuple(), NamedTuple())
```
avik-pal, line 112: I think there is a disconnect between these functions and the original goal. The idea of having these properties in the type, would be to simplify user code. So a stateless layer can define `(::Layer)(x, ps)`. With this the user is still forced to define the original 3 argument api.

src/core.jl
```diff
@@ -91,36 +132,68 @@ end
...
+              AbstractExplicitLayer{hasparams, hasstate} end
 
 function initialparameters(rng::AbstractRNG,
                            l::AbstractExplicitContainerLayer{layers}) where {layers}
     length(layers) == 1 && return initialparameters(rng, getfield(l, layers[1]))
     return NamedTuple{layers}(initialparameters.(rng, getfield.((l,), layers)))
 end
 
+function initialparameters(::AbstractRNG,
+                           ::AbstractExplicitContainerLayer{layers, false, hasstate}) where {
```
avik-pal, line 154: unused `hasstate`

src/core.jl
```diff
@@ -91,36 +132,68 @@ end
...
+                                                                                          hasparams
+                                                                                          }
+    return NamedTuple()
+end
+
 function parameterlength(l::AbstractExplicitContainerLayer{layers}) where {layers}
     return sum(parameterlength, getfield.((l,), layers))
 end
 
+function parameterlength(::AbstractExplicitContainerLayer{layers, false, hasstate}) where {
```
avik-pal, line 179: unused `hasstate`

avik-pal (changes requested):
Made a first pass over the PR.

avik-pal:
Also if we are to simplify user code, we will have to define functions like `(l::AbstractExplicitLayer{false}(x, ps, st) = l(x, st)`,  `(l::AbstractExplicitLayer{hasparams, false}(x, ps, st) = l(x, ps), st`, `(l::AbstractExplicitLayer{false, false}(x, ps, st) = l(x), st`

avik-pal:
Also tagging @chrisrackauckas, @touchesir for thoughts on this

avik-pal:
>  One would need to apply a logical OR to the hasparams / hasstate of each element of the NamedTuple, which sounds like it has to be done dynamically since we don't even know its length.

That information is available during compilation, I don't think it will lead to a dynamic dispatch. Worst case if the compiler can't do that, we can just use a generated function. Also, I wouldn't care too much about it since it is never going to appear in a performance-critical section of the code.

ToucheSir:
I don't see any scenario where you'd need a generated function either, as the aforementioned bool type logic happens at construction. Have you considered whether it makes sense to add type aliases for some of the more common combinations?

ChrisRackauckas:
I'm not sure I like the high level call changes with this. There's something simple about always needing `f(x,p,st)`. If you start being able to drop some in some cases, then a lot of docs will be like "oh yeah it's just `f(x,p)`" (most SciML examples), but then the user goes to put in a BatchNorm and suddenly their code throws an odd error which they don't understand. I think there's simplicity in having only one calling convention.

src/core.jl
gdalle: Fixed in [37f4851](https://github.com/avik-pal/Lux.jl/pull/84/commits/37f48519a731897b9de5771ccb714ff0eea2e013)

src/core.jl
gdalle: Done in [37f4851](https://github.com/avik-pal/Lux.jl/pull/84/commits/37f48519a731897b9de5771ccb714ff0eea2e013)

src/core.jl
gdalle: Fixed in [37f4851](https://github.com/avik-pal/Lux.jl/pull/84/commits/37f48519a731897b9de5771ccb714ff0eea2e013)

src/core.jl
gdalle: Fixed in [37f4851](https://github.com/avik-pal/Lux.jl/pull/84/commits/37f48519a731897b9de5771ccb714ff0eea2e013)

src/core.jl
gdalle: Fixed in [37f4851](https://github.com/avik-pal/Lux.jl/pull/84/commits/37f48519a731897b9de5771ccb714ff0eea2e013)

src/core.jl
gdalle: Fixed in [37f4851](https://github.com/avik-pal/Lux.jl/pull/84/commits/37f48519a731897b9de5771ccb714ff0eea2e013)

gdalle:
> I'm not sure I like the high level call changes with this. 
@ChrisRackauckas I think that depending on your use of the library, those changes might make a lot of sense. For instance, the notion of layer state is not obvious to beginners, and when they don't need it (eg. for a simple CNN without dropout), carrying it along makes the code heavier and less readable.
As mentioned by @avik-pal in #83, it's mostly about quality of life, and to help people get started with Lux.jl. Of course, as soon as you get more familiar with the syntax, you will start adding a state everywhere. But to ease first contact with the code, leaving the state aside for simple networks is a clear bonus to me.

gdalle:
> That information is available during compilation, I don't think it will lead to a dynamic dispatch. Worst case if the compiler can't do that, we can just use a generated function. Also, I wouldn't care too much about it since it is never going to appear in a performance-critical section of the code.

I tried defining the parameters as follows:
```julia
P = maximum(hasparams, layers)
S = maximum(hasstate, layers)
```
but for some reason it messed up type inference for the `NamedTuple` layer field, which made all the tests fail due to the generated functions that are used to apply eg. `Parallel`.

I'll come back to that later.

gdalle:
> Also, I wouldn't care too much about it since it is never going to appear in a performance-critical section of the code.

I'm not sure I agree. There are cases where you want to create a neural network inside a hot loop, and faulty type inference at construction can mess up everything that follows.

ChrisRackauckas:
As I mentioned, no example I have written (which now is like 30 tutorials or so that use Lux) would require state. So the SciML uses of the library would definitely fall into the category that you are trying to target. But even with that, I don't think it's a good idea because this is so pervasive that I think it will make the cases with state confusing. If no tutorials ever include the state, then how are you ever expected to know that it exists? What I foresee happening is that there will be some layers that users will get confused by, like BatchNorm will just throw "mysterious" errors, when in reality the mystery is that the tutorials are all focused on a special case. 

Now this is somewhat hidden in some cases like NeuralPDE and DiffEqFlux  since in those cases users are rarely doing the call themselves. But SciMLSensitivity will definitely have this as an extra documentation issue, and either the solution will be to change our tutorials to showcase state in what will then be a non-standard way (i.e. include state even when it's nothing), always use layers with state in tutorials as a matter of training for the most "robust" form, or just ignoring it as a Lux issue and answering a bunch of issues with "oh, that's a Lux thing. If you use xxx then you have `st` and ...". I hate doing the latter because it quickly leads to 20 emails to answer everyday with the same thing, and by now I've learned that styles which require answering people's questions are just not scalable.

gdalle:
> What I foresee happening is that there will be some layers that users will get confused by, like BatchNorm will just throw "mysterious" errors

I also think that this is a foreseeable problem, but one that can be fixed with explicit error messages on the Lux.jl side when the wrong methods are called. Something like
```
Your neural network contains a layer that relies on a state variable. You probably tried to call it with `network(x, ps)`, but instead you need to use `network(x, ps, st)`. See [...] in the documentation for more details.
```

Anyway, I only started working on this because @avik-pal agreed it would be a welcome feature. If he ends up agreeing it's not, then I'll simply close the PR

gdalle:
Another justification for this feature would be easing the transition from Flux.jl to Lux.jl. Since you announced that SciML would be using Lux.jl, I've been convinced that this transition is a good thing. And to get people hooked to Lux.jl, the fewer visible differences with Flux.jl, the better. Then, when models get more complex and they're in too deep, we can drop the bomb with the additional state argument and they won't be able to escape :sweat_smile:

ChrisRackauckas:
The transition is pretty much done BTW, so check out

- https://sensitivity.sciml.ai/dev/
- https://diffeqflux.sciml.ai/dev/
- https://neuralpde.sciml.ai/dev/

which all bumped majors and removed `FastChain` from everything (well, DiffEqFlux is keeping the depwarns around for awhile to help the updates). 🎉 

Almost everything is Lux now (https://github.com/SciML/NeuralPDE.jl/issues/533 and https://github.com/SciML/NeuralPDE.jl/issues/533 pretty much sealed the deal), except a notable issue with Lux on GPUs (due to ComponentArrays not broadcasting on GPUs correctly). There is an on-going thread about if we're happy with the new coding styles these updates gave, so I encourage you to chime in (https://julialang.slack.com/archives/CN04R7WKE/p1657113244013009). 

Indeed, one of the things that was missing from `FastChain` was state, so when doing these updates we had to add state in. And right now it doesn't look nice, it's using `global`s, which Chris Elrod wasn't too keen on, and there are some `@ignore_derivatives` points which seem to be correct for every case I discussed with @avik-pal though I do wonder whether derivatives w.r.t. state are something that may be required for something (and would love to hear if there is a case). For some things like `NeuralODE` it would be good to disallow models that have some form of state that we know would be incorrect (for example, LSTMs in a `NeuralODE` are bad because the neural ODE doesn't step forward, and so the adaptivity means an LSTM-defined neural ODE does not have a well-defined solution if it rejects steps!), but disallowing all models with state goes too far (normalization layers for example should be fine in the theory that the state is fixed for each ODE solve).

However, even with all of those new issues arising because of state, I think we need to just fine what coding behaviors and habits do it well. For example, always make a callable struct to enclose the state in an ODE definition? We're still working it out, but I think it would be best to find what the nice styles are here and use it everywhere. Otherwise it might just be one "advanced" example that shows it, with other examples showing "worse" coding styles, or even worse, coding styles that have some unnecessary complexity when no state is there (but done anyways for consistency with the cases where it is?).

That said, I will be willing to give it a try since it would simplify docs and such, though I am worried that in all tutorials of these 3 libraries, we would just remove state to match this PR. And if we do that, 😅 state examples will seem like magic. A good error message may be enough though.

gdalle:
> The transition is pretty much done BTW

I know the transition is done for SciML (kudos on the massive change!), but there's lots of other people using Flux out there that we could still convince to switch :wink: 

> That said, I will be willing to give it a try since it would simplify docs and such, though I am worried that in all tutorials of these 3 libraries, we would just remove state to match this PR.

I'm not necessarily advocating for the `layer(x, ps)` to become the default (recommended) syntax, as a matter of fact I'm not sure it should.
And as I stated in [this comment](https://github.com/avik-pal/Lux.jl/pull/84#discussion_r917362838), I think that for people who _define_ new layers, sticking to the three-argument syntax makes more sense.
The change I'm proposing is mostly for people who _use_ pre-defined stateless layers, to which we can give additional options.

avik-pal:
I kind of agree with Chris here that having a consistent API, even if it is not beginner-friendly, might be the way to go. Flux is great in that you can pretty much throw anything its way, and it will work, but for Lux, that is not the case. It relies on a pretty strict convention on how you should be writing your julia/zygote code. 

Originally I was thinking about how to make it easier for people writing custom layers, in which case managing states is quite cumbersome and, in most cases, can be auto-generated. See #46, which kind of already does that (my only concern with merging that PR even though its nearly complete, users will not check the generated code and slap it around places where it might not work correctly)

Also in my experience, it doesn't matter if you show users the "correct" way to do something (especially after introducing a partially correct easier version). They tend to gravitate towards the easier (and in this particular case) the wrong way to do things. Anecdotally, flax which also operates similarly to lux (actually the other way around), introduces a non-state managed API first. And even after two months of using it, I still can't remember how to handle states properly.

gdalle:
Okay, so do we just close this and move on?

gdalle:
I must say the discussion also swayed me towards keeping the state around

avik-pal:
> Okay, so do we just close this and move on?

I think yeah. Also, I should have tagged people in the other issue before you put all this effort in. My bad.

PR #88, opened 2022-07-16
Update comparison section in overview.md

ToucheSir:
Mostly tweaks to verbiage and some updated links. I suspect a good chunk of this was written pre Flux 0.13 :)


avik-pal (approved):

avik-pal:
Yeah that part is from when it was called EFL :)

Issue #94, opened 2022-07-20
Basic example from Migrating from Flux to Lux is broken || normalization issue


gabrevaya:
The Lux version of the first exemplre from in `Migrating from Flux to Lux`  from the docs is broken. The reason is that the input `x` is a `Matrix{Float64}`, but when setting up the model `ps, st = Lux.setup(rng, model)` the parameters are `Matrix{Float32}` by default, and `normalization` function requires the eltypes of all their arguments to be the same. A quick fix for the example in the docs to work is just to initialize `x` as with `Float32`s: `x = randn(rng, Float32, 2, 4)`. However I think it would be good the fix the normalization issue eventually. Also, it would be nice for keeping that line of code between Flux and Lux the same. Or we could make `x = randn(rng, Float32, 2, 4)` for both Lux and Flux.

For fixing the normalization issue, an some options could be to add a method with a different parametrized type for the first argument (which is `x`) or try to promote/convert some of the types if they are not the same. What do you think it would be the best way to handle this?

<details>
  <summary>Code and error message</summary>

  ```julia
using Lux, Random, NNlib, Zygote

model = Chain(Dense(2 => 4), BatchNorm(4, relu), Dense(4 => 2))
rng = Random.default_rng()
x = randn(rng, 2, 4)
ps, st = Lux.setup(rng, model)
model(x, ps, st)
```

```julia
ERROR: MethodError: no method matching normalization(::Matrix{Float64}, ::Vector{Float32}, ::Vector{Float32}, ::Vector{Float32}, ::Vector{Float32}, ::typeof(relu), ::Vector{Int64}, ::Val{true}, ::Float32, ::Float32)
Closest candidates are:
  normalization(::AbstractArray{T, N}, ::Union{Nothing, AbstractVector{T}}, ::Union{Nothing, AbstractVector{T}}, ::Union{Nothing, AbstractVector{T}}, ::Union{Nothing, AbstractVector{T}}, ::Any, ::Any, ::Val) where {T, N} at ~/.julia/packages/Lux/lEqCI/src/nnlib.jl:31
  normalization(::AbstractArray{T, N}, ::Union{Nothing, AbstractVector{T}}, ::Union{Nothing, AbstractVector{T}}, ::Union{Nothing, AbstractVector{T}}, ::Union{Nothing, AbstractVector{T}}, ::Any, ::Any, ::Val, ::T) where {T, N} at ~/.julia/packages/Lux/lEqCI/src/nnlib.jl:31
[...]

PR #97, opened 2022-07-21
making x Float32 in migrate from Flux example

gabrevaya:
Fix the basic example from `Migrating from Flux to Lux` in the docs reported in https://github.com/avik-pal/Lux.jl/issues/94.


avik-pal (approved):
Thanks

Issue #100, opened 2022-07-23
RNN and LSTM break when using GPU


gabrevaya:
Bellow you can find a MWE with `RNNCell`. It is the same for `LSTMCell`.

```julia
using Lux, Random, CUDA

rnn = RNNCell(2 => 8)
rng = Random.default_rng()
Random.seed!(rng, 0)
ps, st = Lux.setup(rng, rnn) .|> gpu
x = rand(Float32, 2, 4, 10) |> gpu
rnn(view(x, :, 1, :), ps, st)
```

```julia
ERROR: ArgumentError: cannot take the CPU address of a CuArray{Float32, 2, CUDA.Mem.DeviceBuffer}
Stacktrace:
  [1] unsafe_convert(#unused#::Type{Ptr{Float32}}, x::CuArray{Float32, 2, CUDA.Mem.DeviceBuffer})
    @ CUDA ~/.julia/packages/CUDA/DfvRa/src/array.jl:319
  [2] gemm!(transA::Char, transB::Char, alpha::Float32, A::CuArray{Float32, 2, CUDA.Mem.DeviceBuffer}, B::Matrix{Float32}, beta::Float32, C::Matrix{Float32})
    @ LinearAlgebra.BLAS /network/scratch/a/abrevayg/julia-1.8.0-rc3/share/julia/stdlib/v1.8/LinearAlgebra/src/blas.jl:1514
  [3] gemm_wrapper!(C::Matrix{Float32}, tA::Char, tB::Char, A::CuArray{Float32, 2, CUDA.Mem.DeviceBuffer}, B::Matrix{Float32}, _add::LinearAlgebra.MulAddMul{true, true, Bool, Bool})
    @ LinearAlgebra /network/scratch/a/abrevayg/julia-1.8.0-rc3/share/julia/stdlib/v1.8/LinearAlgebra/src/matmul.jl:674
  [4] mul!
    @ /network/scratch/a/abrevayg/julia-1.8.0-rc3/share/julia/stdlib/v1.8/LinearAlgebra/src/matmul.jl:161 [inlined]
  [5] mul!
    @ /network/scratch/a/abrevayg/julia-1.8.0-rc3/share/julia/stdlib/v1.8/LinearAlgebra/src/matmul.jl:276 [inlined]
  [6] *
    @ /network/scratch/a/abrevayg/julia-1.8.0-rc3/share/julia/stdlib/v1.8/LinearAlgebra/src/matmul.jl:148 [inlined]
[...]

PR #101, opened 2022-07-23
add init_hidden_state function

gabrevaya:
This is a possible simple patch for https://github.com/avik-pal/Lux.jl/issues/100, although I don't know if it is the best.

Initially, I tried to dispatch over `CuArrays` but then I noticed that their views weren't caught by that, so eventually I dispatched over `CUDA.StridedSubCuArray`, which works but I don't know if there is a better solution.

Passing the `init_state` through `gpu` would work for any `init_state`, but maybe in the case of the default state initialization, which is zeros, would it be better to initialize it with CUDA.zeros to save some allocations?

Also, I'm not sure if I have to do something with `rng` (like `replicate` it and `@set!` it) or if just passing it to `init_state` is enough.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -194,11 +194,16 @@ end
 @inline _gate(h::Int, n::Int) = (1:h) .+ h * (n - 1)
 @inline _gate(x::AbstractVector, h::Int, n::Int) = view(x, _gate(h, n))
 @inline _gate(x::AbstractMatrix, h::Int, n::Int) = view(x, _gate(h, n), :)
-@inline init_hidden_state(rng::AbstractRNG, rnn, x::AbstractMatrix) =
-    rnn.init_state(rng, rnn.out_dims, size(x, 2))
-@inline init_hidden_state(rng::AbstractRNG, rnn, x::Union{CUDA.StridedSubCuArray, CuArray}) =
-    rnn.init_state(rng, rnn.out_dims, size(x, 2)) |> gpu
-    
+
+@inline function init_hidden_state(rng::AbstractRNG, rnn, x::AbstractMatrix)
+    return rnn.init_state(rng, rnn.out_dims, size(x, 2))
+end
+
+@inline function init_hidden_state(rng::AbstractRNG, rnn,
+                                   x::Union{CUDA.StridedSubCuArray, CuArray})
+    return rnn.init_state(rng, rnn.out_dims, size(x, 2)) |> gpu
+end
+
 """
     multigate(x::AbstractArray, ::Val{N})
 
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -195,11 +195,11 @@ end
 @inline _gate(x::AbstractVector, h::Int, n::Int) = view(x, _gate(h, n))
 @inline _gate(x::AbstractMatrix, h::Int, n::Int) = view(x, _gate(h, n), :)
 
-@inline function init_hidden_state(rng::AbstractRNG, rnn, x::AbstractMatrix)
+@inline function _init_hidden_state(rng::AbstractRNG, rnn, x::AbstractMatrix)
     return rnn.init_state(rng, rnn.out_dims, size(x, 2))
 end
 
-@inline function init_hidden_state(rng::AbstractRNG, rnn,
+@inline function _init_hidden_state(rng::AbstractRNG, rnn,
                                    x::Union{CUDA.StridedSubCuArray, CuArray})
     return rnn.init_state(rng, rnn.out_dims, size(x, 2)) |> gpu
 end
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -201,7 +201,7 @@ end
 
 @inline function _init_hidden_state(rng::AbstractRNG, rnn,
                                     x::Union{CUDA.StridedSubCuArray, CuArray})
-    return rnn.init_state(rng, rnn.out_dims, size(x, 2)) |> gpu
+    return CuArray(rnn.init_state(rng, rnn.out_dims, size(x, 2)))
 end
 
 """
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -194,6 +194,8 @@ end
 @inline _gate(h::Int, n::Int) = (1:h) .+ h * (n - 1)
 @inline _gate(x::AbstractVector, h::Int, n::Int) = view(x, _gate(h, n))
 @inline _gate(x::AbstractMatrix, h::Int, n::Int) = view(x, _gate(h, n), :)
+@inline init_hidden_state(rng::AbstractRNG, rnn, x::AbstractMatrix) = rnn.init_state(rng, rnn.out_dims, size(x, 2))
+@inline init_hidden_state(rng::AbstractRNG, rnn, x::CUDA.StridedSubCuArray) = rnn.init_state(rng, rnn.out_dims, size(x, 2)) |> gpu
 
 """
     multigate(x::AbstractArray, ::Val{N})
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -194,9 +194,11 @@ end
 @inline _gate(h::Int, n::Int) = (1:h) .+ h * (n - 1)
 @inline _gate(x::AbstractVector, h::Int, n::Int) = view(x, _gate(h, n))
 @inline _gate(x::AbstractMatrix, h::Int, n::Int) = view(x, _gate(h, n), :)
-@inline init_hidden_state(rng::AbstractRNG, rnn, x::AbstractMatrix) = rnn.init_state(rng, rnn.out_dims, size(x, 2))
-@inline init_hidden_state(rng::AbstractRNG, rnn, x::CUDA.StridedSubCuArray) = rnn.init_state(rng, rnn.out_dims, size(x, 2)) |> gpu
-
+@inline init_hidden_state(rng::AbstractRNG, rnn, x::AbstractMatrix) =
+    rnn.init_state(rng, rnn.out_dims, size(x, 2))
+@inline init_hidden_state(rng::AbstractRNG, rnn, x::Union{CUDA.StridedSubCuArray, CuArray}) =
+    rnn.init_state(rng, rnn.out_dims, size(x, 2)) |> gpu
+    
 """
     multigate(x::AbstractArray, ::Val{N})
 
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -200,7 +200,7 @@ end
 end
 
 @inline function _init_hidden_state(rng::AbstractRNG, rnn,
-                                   x::Union{CUDA.StridedSubCuArray, CuArray})
+                                    x::Union{CUDA.StridedSubCuArray, CuArray})
     return rnn.init_state(rng, rnn.out_dims, size(x, 2)) |> gpu
 end
 
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Lux"
 uuid = "b2108857-7c20-44ae-9111-449ecde12c47"
 authors = ["Avik Pal <avikpal@mit.edu> and contributors"]
-version = "0.4.9"
+version = "0.4.10"
 
 [deps]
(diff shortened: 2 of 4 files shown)


src/utils.jl
```diff
@@ -194,6 +194,8 @@ end
 @inline _gate(h::Int, n::Int) = (1:h) .+ h * (n - 1)
 @inline _gate(x::AbstractVector, h::Int, n::Int) = view(x, _gate(h, n))
 @inline _gate(x::AbstractMatrix, h::Int, n::Int) = view(x, _gate(h, n), :)
+@inline init_hidden_state(rng::AbstractRNG, rnn, x::AbstractMatrix) = rnn.init_state(rng, rnn.out_dims, size(x, 2))
+@inline init_hidden_state(rng::AbstractRNG, rnn, x::CUDA.StridedSubCuArray) = rnn.init_state(rng, rnn.out_dims, size(x, 2)) |> gpu
```
avik-pal, line 198:
```suggestion
@inline init_hidden_state(rng::AbstractRNG, rnn, x::CUDA.CuMatrix) = rnn.init_state(rng, rnn.out_dims, size(x, 2)) |> gpu
```

gabrevaya:
I changed  `x::CUDA.StridedSubCuArray` for `x::Union{CUDA.StridedSubCuArray, CuArray}` to make it compatible with passing just a slice of the array `x[:, 1, :]` instead of a view of it `view(x, :, 1, :)`.

I also corrected the format. I have some doubts about it. I run the JuliaFormatter (with the project specific .JuliaFormatter.toml) and many functions from utils.jl were modified. Just in case for some reason you prefer it to be the way it was (skipping some rules from JuliaFormatter) I left all the rest unchanged and only formated the new init_hidden_state. Should we apply the JuliaFormatter suggested changes to all the rest? Also, I noticed that the JuliaFormatter always remove the spaces between types between the types declarations, for example: `Union{ComponentArray, NamedTuple}` -> `Union{ComponentArray,NamedTuple}`. Since I saw that all the types in the package are separated by a space, I put the space back to `Union{CUDA.StridedSubCuArray, CuArray}`. However, if the idea is to follow the JuliaFormatter style we should try to make it compatible (otherwise, mention this in the docs' Style Guide). If you'd like I can open a new issue about this.

gabrevaya:
Actually, for the first line of

```julia
@inline init_hidden_state(rng::AbstractRNG, rnn, x::Union{CUDA.StridedSubCuArray, CuArray}) =
    rnn.init_state(rng, rnn.out_dims, size(x, 2)) |> gpu
```
the length is exceeded by 1 character. The original suggestion by JuliaFormatter was without the spaces between the types and then the length was good.  Is it preferable to have the space and exceed in 1 character, without the space and abide by the length limit, or break the line?

avik-pal:
Use full `function` syntax instead of multiline

src/utils.jl
```diff
@@ -195,6 +195,15 @@ end
 @inline _gate(x::AbstractVector, h::Int, n::Int) = view(x, _gate(h, n))
 @inline _gate(x::AbstractMatrix, h::Int, n::Int) = view(x, _gate(h, n), :)
 
+@inline function init_hidden_state(rng::AbstractRNG, rnn, x::AbstractMatrix)
```
avik-pal, line 198: Lets make these `_init_hidden_state` since we are not exposing them to the users

gabrevaya:
Should I add a test for `_init_hidden_state`?

avik-pal:
Yeah that you be great. Also, bump the patch version. I will tag a release once merged

gabrevaya:
Why is the coverage now worse than before adding the tests?? Does it have to do with the version bump or did I do something wrong with the tests?

avik-pal:
The builds failed

gabrevaya:
Do you know why they failed? Should I do anything?

avik-pal:
https://github.com/avik-pal/Lux.jl/runs/7484682987?check_suite_focus=true#step:6:183

avik-pal:
The formatting seems off. Run `format` and we should be good to merge

gabrevaya:
Oh, sorry for that. I'll try to fix it now. With 'run `format`' you mean running the JuliaFormatter? If so, should I keep all the changes that it makes? As I mentioned [above](https://github.com/avik-pal/Lux.jl/pull/101#issuecomment-1193152708), when I ran it on layers/utils.jl, there were changes in many lines other than the newly added by this PR, and some of them had a different style than the rest of the package.

avik-pal:
Just to confirm are you using format from the REPL and in the root dir of the project? I know that running from vscode doesn't respect the format guidelines rn.

gabrevaya:
I was running it from the REPL integrated in VS Code, I'll try directly from the terminal.

gabrevaya:
It is throwing this error:
<details>
  <summary> Error log </summary>

```
julia -e 'using Pkg; Pkg.add(PackageSpec(name="JuliaFormatter"))'
```

```julia
    Updating registry at `~/.julia/registries/General.toml`
   Resolving package versions...
    Updating `~/.julia/environments/v1.8/Project.toml`
⌃ [98e50ef6] + JuliaFormatter v1.0.6
  No Changes to `~/.julia/environments/v1.8/Manifest.toml`
```

```
julia -e 'using JuliaFormatter; format(".")'
```

```julia
ERROR: Parsing error for input occurred on line 19, offset: 56
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:35
  [2] format_text(text::String, style::SciMLStyle, opts::JuliaFormatter.Options)
    @ JuliaFormatter ~/.julia/packages/JuliaFormatter/D5Gl0/src/JuliaFormatter.jl:603
  [3] format_text(text::String, style::SciMLStyle; maxiters::Int64, kwargs::Base.Pairs{Symbol, Integer, NTuple{8, Symbol}, NamedTuple{(:separate_kwargs_with_semicolon, :format_docstrings, :always_for_in, :join_lines_based_on_source, :margin, :whitespace_in_kwargs, :always_use_return, :indent), Tuple{Bool, Bool, Bool, Bool, Int64, Bool, Bool, Int64}}})
    @ JuliaFormatter ~/.julia/packages/JuliaFormatter/D5Gl0/src/JuliaFormatter.jl:588
  [4] #format_text#226
    @ ~/.julia/packages/JuliaFormatter/D5Gl0/src/JuliaFormatter.jl:574 [inlined]
  [5] format_file(filename::String; overwrite::Bool, verbose::Bool, format_markdown::Bool, format_options::Base.Pairs{Symbol, Any, NTuple{9, Symbol}, NamedTuple{(:separate_kwargs_with_semicolon, :format_docstrings, :always_for_in, :join_lines_based_on_source, :style, :margin, :whitespace_in_kwargs, :always_use_return, :indent), Tuple{Bool, Bool, Bool, Bool, SciMLStyle, Int64, Bool, Bool, Int64}}})
    @ JuliaFormatter ~/.julia/packages/JuliaFormatter/D5Gl0/src/JuliaFormatter.jl:737
  [6] (::JuliaFormatter.var"#233#236"{Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}}})(formatted_file::Bool, file::String)
    @ JuliaFormatter ~/.julia/packages/JuliaFormatter/D5Gl0/src/JuliaFormatter.jl:821
  [7] BottomRF
    @ ./reduce.jl:81 [inlined]
  [8] _foldl_impl(op::Base.BottomRF{JuliaFormatter.var"#233#236"{Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}}}}, init::Bool, itr::Vector{String})
    @ Base ./reduce.jl:62
  [9] foldl_impl
    @ ./reduce.jl:48 [inlined]
 [10] mapfoldl_impl
    @ ./reduce.jl:44 [inlined]
 [11] _mapreduce_dim
    @ ./reducedim.jl:362 [inlined]
 [12] #mapreduce#765
    @ ./reducedim.jl:357 [inlined]
 [13] #reduce#767
    @ ./reducedim.jl:406 [inlined]
 [14] (::JuliaFormatter.var"#232#235"{Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}}})(formatted_path::Bool, dir_branch::Tuple{String, Vector{String}, Vector{String}})
    @ JuliaFormatter ~/.julia/packages/JuliaFormatter/D5Gl0/src/JuliaFormatter.jl:809
 [15] BottomRF
    @ ./reduce.jl:81 [inlined]
 [16] _foldl_impl(op::Base.BottomRF{JuliaFormatter.var"#232#235"{Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}}}}, init::Bool, itr::Channel{Tuple{String, Vector{String}, Vector{String}}})
    @ Base ./reduce.jl:58
 [17] foldl_impl(op::Base.BottomRF{JuliaFormatter.var"#232#235"{Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}}}}, nt::Bool, itr::Channel{Tuple{String, Vector{String}, Vector{String}}})
    @ Base ./reduce.jl:48
 [18] mapfoldl_impl
    @ ./reduce.jl:44 [inlined]
 [19] #mapfoldl#259
    @ ./reduce.jl:162 [inlined]
 [20] #mapreduce#263
    @ ./reduce.jl:294 [inlined]
 [21] #reduce#265
    @ ./reduce.jl:475 [inlined]
 [22] format(paths::Tuple{String}; options::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
    @ JuliaFormatter ~/.julia/packages/JuliaFormatter/D5Gl0/src/JuliaFormatter.jl:806
 [23] format
    @ ~/.julia/packages/JuliaFormatter/D5Gl0/src/JuliaFormatter.jl:779 [inlined]
 [24] #format#237
    @ ~/.julia/packages/JuliaFormatter/D5Gl0/src/JuliaFormatter.jl:831 [inlined]
 [25] format(path::String)
    @ JuliaFormatter ~/.julia/packages/JuliaFormatter/D5Gl0/src/JuliaFormatter.jl:831
 [26] top-level scope
    @ none:1
```
</details>

gabrevaya:
Now it worked. I had two uncommited files there that were causing the problems.

gabrevaya:
Wait a min before merging. I think I might have caused some type instability. I'm checking it now.

gabrevaya:
Yes, the problem is the `gpu` function which is not type stable. Do you have any alternative solution?

gabrevaya:
Maybe we can just replace `gpu` for `CuArray`
```julia
@inline function _init_hidden_state(rng::AbstractRNG, rnn,
                                   x::Union{CUDA.StridedSubCuArray, CuArray})
    return rnn.init_state(rng, rnn.out_dims, size(x, 2)) |> CuArray
end
```
What do you think?

gabrevaya:
Or should we use `CUDA.cu` or `adapt_storage`?

gabrevaya:
Actually, using `CUDA.cu` is still type unstable, so maybe we should just use `CuArray`. I'll wait for your confirmation before commiting.

src/utils.jl
```diff
@@ -195,6 +195,15 @@ end
 @inline _gate(x::AbstractVector, h::Int, n::Int) = view(x, _gate(h, n))
 @inline _gate(x::AbstractMatrix, h::Int, n::Int) = view(x, _gate(h, n), :)
 
+@inline function _init_hidden_state(rng::AbstractRNG, rnn, x::AbstractMatrix)
+    return rnn.init_state(rng, rnn.out_dims, size(x, 2))
+end
+
+@inline function _init_hidden_state(rng::AbstractRNG, rnn,
+                                    x::Union{CUDA.StridedSubCuArray, CuArray})
+    return rnn.init_state(rng, rnn.out_dims, size(x, 2)) |> gpu
```
avik-pal, line 204: I agree lets change this to 
```suggestion
    return CuArray(rnn.init_state(rng, rnn.out_dims, size(x, 2)))
```

avik-pal (approved):

PR #117, opened 2022-08-01
Trainable hidden states

lungd:
What do you think about a `trainable_hidden_states` flag similar to `bias`?
With trainable_hidden_state=true, ps contains a hidden_state vector which gets repeated when calling the layer without hidden_states.


avik-pal:
Thanks. This looks like an interesting feature and I can see its utility of it. Can you reference some paper that uses this?

lungd:
I am not aware of any concrete work/model which uses trainable initial hidden states but I have the feeling, zero or randomly initialized vectors are more common.
 
I didn't look for that long but maybe as ref

[Learning Precise Timing with LSTM Recurrent Networks - 5.1](https://www.jmlr.org/papers/volume3/gers02a/gers02a.pdf):
```
At the beginning of each stream cell states and gate activations are initialized to zero. 
This initial state is almost always quite far from the corresponding state in the same phase of later periods in the stream. 
After few consecutive periods, cell states and gate activations of successful networks tend to settle to very stable,  phase-specific values, which are typically quite different from the corresponding values in the first period. 
This suggests that the initial state of the network should be learned as well,as proposed by Forcada and Carrasco (1995),Bulsari and Sax ́en (1995),instead of arbitrarily initializing it to zero
```

[r2rt Blogpost](https://r2rt.com/non-zero-initial-states-for-recurrent-neural-networks.html):
```
From the above experiment we make the following observations:
    * All non-zero state intializations sped up training and improved generalization.
    * Training the initial state as a variable was more effective than using a noisy zero-mean initial state.
    * Adding noise to a variable initial state provided only marginal benefit.
```

avik-pal:
Yeah, I guessed it is more like "we all know about this," but no one wrote it up formally. We can go ahead with this feature.

lungd:
Should `show()` display `train_state=true` if it is set to true or would it even be better to have trainable hidden states by default?

avik-pal:
The default should be no trainable states from a consistency with most frameworks and non breaking standpoint.

Yeah let's add `train_state=true` to show.

avik-pal (approved):

avik-pal:
Version bump, and we are ready to merge.

Issue #118, opened 2022-08-03
Make it easier to pass empty state `st = (;)`


gdalle:
Hi there!
I am writing code with stateless neural networks, and I would like to be able to pass `st = (;)` (empty `NamedTuple`) everywhere instead of remembering the state from the `Lux.setup` funtion.
Judging by the code in `layers/basic.jl`, it should work with every layer... except `Chain` because of [this line](https://github.com/avik-pal/Lux.jl/blob/9924210398817d285f582a694677d87b72c12907/src/layers/basic.jl#L507), where `st` is annotated as a `NamedTuple{fields}` instead of a `NamedTuple`. Is there a specific reason for this?

avik-pal:
Actually, the other ones should be annotated with `fields`. Also I don't think it will do what you want. It still tries to access the fields of the named tuple, so for an empty namedtuple it should error.

gdalle:
At least the PR passes the tests :shrug: Since that `Chain` was incoherent with the rest, I knew it was either all or nothing on the `NamedTuple` annotation. Just kinda sad it ends up being all instead of nothing :wink:

gdalle:
Do you want a PR that does the change in the reverse direction, and annotates all `st` arguments with `NamedTuple{blabla}`? I think that one would be breaking though

avik-pal:
It's not breaking, `apply*` are not public API. Also in either case it would error, with stricter types IMO it is easier to debug.

> At least the PR passes the tests

Its because the layers use the correct structure. The failure will only occur if we pass an incorrectly structured nt.

lungd:
Speaking of tests, are the tests setup and working properly?
With that [PR](https://github.com/avik-pal/Lux.jl/pull/114/files#diff-b6c250c343270a4d87fe8ae8dc99e882c5a1b7464e516e60b4b28e14856fb202R746) [basic.jl#268](https://github.com/avik-pal/Lux.jl/blob/main/test/layers/basic.jl#L268) should have failed?

avik-pal:
>Speaking of tests, are the tests setup and working properly?

That's just an untested code path.

gdalle:
OK, so let's close this? One day, I'll make a PR to Lux that gets accepted :muscle:

PR #119, opened 2022-08-03
Generic type for st in Chain

gdalle:
Addresses #118


avik-pal:
Closing it as per #118.

PR #120, opened 2022-08-03
Rnn bias deprecation

lungd:
This PR deprecates RNNCell.bias to make it consistent with base.jl
In other ML frameworks it is possible to deactivate the bias of other recurrent layers as well (e.g. LSTM).
I could add a `use_bias` without `bias` to LSTM and GRU but I think in this case a deprecated `bias` would be a better choice so you get the same behavior for all the layers (i.e. you can create RNNCells and LSTMs with `bias`, which prints a warning)

### DIFF
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -49,6 +49,7 @@ end
 function RNNCell((in_dims, out_dims)::Pair{<:Int, <:Int}, activation=tanh;
                  use_bias::Bool=true, bias::Union{Missing, Bool}=missing, init_bias=zeros32,
                  init_weight=glorot_uniform, init_state=ones32)
+    # Deprecated Functionality (Remove in v0.5)
     if !ismissing(bias)
         Base.depwarn("`bias` argument to `RNNCell` has been deprecated and will be removed" *
                      " in v0.5. Use `use_bias` kwarg instead.", :RNNCell)
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -79,8 +79,7 @@ function initialstates(rng::AbstractRNG, ::RNNCell)
     return (rng=replicate(rng),)
 end
 
-function (rnn::RNNCell)(x::AbstractMatrix, ps::Union{ComponentArray, NamedTuple},
-                        st::NamedTuple)
+function (rnn::RNNCell)(x::AbstractMatrix, ps, st::NamedTuple)
     rng = replicate(st.rng)
     @set! st.rng = rng
     hidden_state = _init_hidden_state(rng, rnn, x)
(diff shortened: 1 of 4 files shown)


avik-pal:
> I could add a use_bias without bias to LSTM and GRU but I think in this case a deprecated bias would be a better choice so you get the same behavior for all the layers (i.e. you can create RNNCells and LSTMs with bias, which prints a warning)

Let's do that in a separate PR. Also go with `use_bias` since users should not be using bias kwarg

lungd:
Waiting for approval

src/layers/recurrent.jl
```diff
@@ -46,17 +46,28 @@ struct RNNCell{bias, A, B, W, S} <: AbstractExplicitLayer
     init_state::S
 end
 
-function RNNCell((in_dims, out_dims)::Pair{<:Int, <:Int}, activation=tanh; bias::Bool=true,
-                 init_bias=zeros32, init_weight=glorot_uniform, init_state=ones32)
-    return RNNCell{bias, typeof(activation), typeof(init_bias), typeof(init_weight),
+function RNNCell((in_dims, out_dims)::Pair{<:Int, <:Int}, activation=tanh;
+                 use_bias::Bool=true, bias::Union{Missing, Bool}=missing, init_bias=zeros32,
+                 init_weight=glorot_uniform, init_state=ones32)
+    if !ismissing(bias)
```
avik-pal, line 52: Can you add a msg like https://lux.csail.mit.edu/dev/devdocs/style_guide/#deprecation so that we don't miss out things while deleting things?

src/layers/recurrent.jl
```diff
@@ -106,10 +117,10 @@ function (rnn::RNNCell{false, typeof(identity)})((x,
     return h_new, st
 end
 
-function Base.show(io::IO, r::RNNCell{bias}) where {bias}
+function Base.show(io::IO, r::RNNCell{use_bias}) where {use_bias}
     print(io, "RNNCell($(r.in_dims) => $(r.out_dims)")
     (r.activation == identity) || print(io, ", $(r.activation)")
-    bias || print(io, ", bias=false")
+    use_bias || print(io, ", use_bias=false")
```
avik-pal, line 123: I think keeping this as `bias` is better. But if we are making this change, we need to do it for all the bias optional layers.

avik-pal (changes requested):
Also, version bump.

src/layers/recurrent.jl
lungd: This was actually an unintended change, I've fixed now

src/layers/recurrent.jl
avik-pal: Add this comment `# Deprecated Functionality (Remove in v0.5)`

lungd: The constructor does already warn about the deprecated kwarg similar to Dense. The testset contains a nested 'Deprecations' testset. I've now added a message about the new `use_bias` kwarg to the changelog.

Do you want me to move the `Base.depwarn` from src/layers/recurrent.jl to src/deprecated.jl?

avik-pal: No everything else is fine. Just add the comment above the `if !dismissing(bias)` like.

lungd: done

avik-pal (approved):

PR #132, opened 2022-08-12
LSTM docs: don't go over first element in sequence twice

visr:
Perhaps something like this would be more robust, though not sure if you'd prefer that for documentation:

```julia
x_init, x_rest = Iterators.peel(eachslice(x, dims=2))
```


avik-pal:
Let's go with the eachslice implementation. It is definitely more robust

visr:
Ok I went with that in cb30ab49b888e82327fe25f9129f13e3dc4c7370.

Let me know if you prefer to take this out, but in 075e9a4de40289fc66ed05ae99e3dde54e7b4085 I added an example LSTM layer definition to the example, as I felt the example is easier to follow this way.

avik-pal (approved):

PR #138, opened 2022-08-22
Add kaiming initialization

YichengDWu:
Directly taken form [Flux](https://github.com/FluxML/Flux.jl/blob/d4f1d816563edd5f953a3fd1ef7dd960d507ed22/src/utils.jl#L212)


avik-pal (changes requested):
Needs tests.

avik-pal (approved):

PR #142, opened 2022-08-31
Change Indentation making it more consistent with the helper packages

avik-pal:


lungd:
What about `Use 4 spaces per indentation level, no tabs.`?
https://github.com/SciML/SciMLStyle#high-level-rules

avik-pal:
I typically refer to the SciMLStyle as a guideline rather than a set of strict rules. 2 spaces kind of look more compact and visually appealing (has a lot of do with me recently having to work on 2 spaces codebases).

I am happy to be convinced as to why not use it.

zsz00:
Use 2 spaces per indentation is too consistent, when code have more lines, poor readability. 
it seam that why not use 1 spaces per indentation making it more consistent. 

I think use 4 spaces per indentation level is good.

lungd:
For me it was the other way round. I was used to see and also use myself 2 spaces per indentation level. That changed mainly due to the SciMLStyle.
I'd say it's definitely a matter of taste and that devs should not be forced to format their code in a certain way, but I think it would be beneficial if all Julia projects follow the same guideline.
I've mentioned the 4 spaces rule just because I know the README contains a SciMLStyle badge.

PR #274, opened 2023-03-03
Some Improvements

avik-pal:
* Remove some of the repeat rules for tracker
* restructure some files
* truncated stacktraces


src/stacktraces.jl
```diff
@@ -0,0 +1,122 @@
...
+
+    fdef = Dict(:name => name, :args => args, :kwargs => kwargs, :body => body,
+                :whereparams => whereparams)
+
+    return eval(MacroTools.combinedef(fdef))
+end
+
+function __get_parameter_count(T::Union{DataType, UnionAll})
+    return length(Base.unwrap_unionall(T).parameters)
+end
```
avik-pal, line 73: @ChrisRackauckas would you accept a macro like this for `TruncatedStacktraces.jl`. Makes life quite easy at least for simple cases.

ChrisRackauckas: yes that would be nice to have.

avik-pal: Added it here https://github.com/SciML/TruncatedStacktraces.jl/pull/7

PR #319, opened 2023-06-07
Added '__apply_activation' to 'RNNCell's

MartinuzziFrancesco:
A follow up of #293 I think

### DIFF
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -267,16 +267,14 @@ const _RNNCellInputType = Tuple{<:AbstractMatrix, Tuple{<:AbstractMatrix}}
 
 function (rnn::RNNCell{true})((x, (hidden_state,))::_RNNCellInputType, ps, st::NamedTuple)
     h_new = ps.weight_ih * x .+ ps.weight_hh * hidden_state .+ ps.bias
-    return (__apply_activation(rnn.activation, h_new),
-        (__apply_activation(rnn.activation, h_new),)),
-    st
+    h_new = __apply_activation(rnn.activation, h_new)
+    return (h_new, (h_new,)), st
 end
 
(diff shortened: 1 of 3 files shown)


src/layers/recurrent.jl
```diff
@@ -266,27 +266,17 @@ end
...
-end
-
-function (rnn::RNNCell{true, TS, typeof(identity)})((x, (hidden_state,))::_RNNCellInputType,
-    ps,
-    st::NamedTuple) where {TS}
     h_new = ps.weight_ih * x .+ ps.weight_hh * hidden_state .+ ps.bias
-    return (h_new, (h_new,)), st
+    return (__apply_activation(rnn.activation, h_new),
+        (__apply_activation(rnn.activation, h_new),)),
+    st
```
avik-pal, line 272:
```suggestion
    h_new = __apply_activation(rnn.activation, h_new)
    return (h_new, (h_new,)), st
```

src/layers/recurrent.jl
```diff
@@ -266,27 +266,17 @@ end
...
-
-function (rnn::RNNCell{false, TS, typeof(identity)})((x, (hs,))::_RNNCellInputType,
-    ps,
-    st::NamedTuple) where {TS}
-    h_new = ps.weight_ih * x .+ ps.weight_hh * hs
-    return (h_new, (h_new,)), st
+    h_new = ps.weight_ih * x .+ ps.weight_hh * hidden_state
+    return (__apply_activation(rnn.activation, h_new),
+        (__apply_activation(rnn.activation, h_new),)),
+    st
```
avik-pal, line 279:
```suggestion
    h_new = __apply_activation(rnn.activation, h_new)
    return (h_new, (h_new,)), st
```

avik-pal (changes requested):

src/layers/recurrent.jl
MartinuzziFrancesco: just pushed the suggested changes

Issue #320, opened 2023-06-08
Adding `AbstractRecurrentCell`


MartinuzziFrancesco:
I was reading through `recurrent.jl` and sqw the [comment](https://github.com/LuxDL/Lux.jl/blob/main/src/layers/recurrent.jl#L1) about adding another subtype for dispatching over `Recurrence` and `StatefulRecurrentCell` more safely. Would something along the lines of 
```julia
abstract type AbstractRecurrentCell{use_bias, train_state} <: 
    AbstractExplicitLayer end

struct Recurrence{R, C <: AbstractRecurrentCell} <: 
    AbstractExplicitContainerLayer{(:cell,)}
    cell::C
end

struct StatefulRecurrentCell{C <: AbstractRecurrentCell} <:
       AbstractExplicitContainerLayer{(:cell,)}
    cell::C
end

struct RNNCell{use_bias, train_state, A, B, W, S} <: AbstractRecurrentCell{use_bias, train_state}
    activation::A
    in_dims::Int
    out_dims::Int
    init_bias::B
    init_weight::W
    init_state::S
end

etcetc...
```
be fine or did you have a more specific design in mind?

avik-pal:
Yes, this is pretty much what I had in mind

MartinuzziFrancesco:
Very nice, I should have some code in place already! I'll open a PR

PR #322, opened 2023-06-08
Added `AbstractRecurrentCell`

MartinuzziFrancesco:
Addressing #320

### DIFF
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -1,5 +1,4 @@
-# TODO(@avik-pal): We can add another subtype `AbstractRecurrentCell` in the type hierarchy
-#                  to make it safer to compose these cells with `Recurrence`
+abstract type AbstractRecurrentCell{use_bias, train_state} <: AbstractExplicitLayer end
 """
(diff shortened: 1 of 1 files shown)


src/layers/recurrent.jl
```diff
@@ -129,10 +128,20 @@ function (r::StatefulRecurrentCell)(x, ps, st::NamedTuple)
     return out, (; cell=st_, carry)
 end
 
-function applyrecurrentcell(l::AbstractExplicitLayer, x, ps, st, carry)
+function applyrecurrentcell(l::AbstractRecurrentCell{use_bias, train_state},
+    x,
+    ps,
+    st,
+    carry) where {use_bias, train_state}
```
avik-pal, line 135:
```suggestion
function applyrecurrentcell(l::AbstractRecurrentCell, x, ps, st, carry)
```

avik-pal (changes requested):
`use_bias` and `train_state` were not being used, so we can remove those.

MartinuzziFrancesco:
gotcha, removed as suggested

PR #491, opened 2024-02-15
Add output lengths for layers

SebastianM-C:
This PR implements the `inputsize` and `outputsize` interface defined in LuxCore in https://github.com/LuxDL/LuxCore.jl/pull/19

I'm not sure if defined this correctly for `ReshapeLayer` and `Scale`.

Also, I'm not sure if the functions can be defined for some of the layers.


src/layers/basic.jl
```diff
@@ -300,6 +305,9 @@ end
 parameterlength(d::Scale{use_bias}) where {use_bias} = (1 + use_bias) * prod(d.dims)
 statelength(d::Scale) = 0
 
+inputsize(d::Scale) = d.dims
+outputsize(d::Scale) = d.dims
```
avik-pal, line 309: Should be tuple?

avik-pal:
Can you also add them to the LuxCore API docs?

avik-pal:
+ Bump the lowerbound for LuxCore

src/layers/basic.jl
SebastianM-C: Oh, right, I thought it already is one.

avik-pal: Ah I see

test/layers/basic_tests.jl
```diff
@@ -9,6 +9,7 @@
             x = randn(rng, 6, 3) |> aType
 
             @test size(layer(x, ps, st)[1]) == (2, 3, 3)
+            @test Lux.outputsize(layer) == (2, 3)
```
SebastianM-C, line 12: I'm not entirely sure if the definition of `outputsize` makes sense for `ReshapeLayer`.

avik-pal: why?

test/layers/basic_tests.jl
```diff
@@ -183,6 +187,9 @@ end
             @test size(first(Lux.apply(layer, randn(10) |> aType, ps, st))) == (10, 5)
             @test size(first(Lux.apply(layer, randn(10, 5, 2) |> aType, ps, st))) ==
                   (10, 5, 2)
+
+            @test LuxCore.inputsize(layer) == (10, 5)
+            @test LuxCore.outputsize(layer) == (10, 5)
```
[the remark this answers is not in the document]
avik-pal: I think outputsize is fine. Do we really need to be able to define `inputsize` for Symbolics? I feel we should return a set of possible sizes for the input.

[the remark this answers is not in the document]
SebastianM-C: No. I defined it only for consistency, but I'm not sure where it would be useful.

test/layers/basic_tests.jl
SebastianM-C: I'm a bit confused about the fact that the size of the output in the test above (and also in general) has a different size compared to `outputsize` because the input also matters and we don't know it from the layer definition.

SebastianM-C:
Should I just implement `outputsize` in this PR? I think `inputsize` is harder to define and I'm not sure if it's even a good idea in the first place. If we think about it in terms of all possible combinations, I'm afraid it could lead to combinatorial explosion (for `ReshapeLayer` you'd basically have to invert `reshape` for arbitrary inputs). But the main thing is that it's not needed for anything (that I know of).

avik-pal:
Yes let's restrict the current PR to `inputsize`

SebastianM-C:
> Yes let's restrict the current PR to `inputsize`

You mean `outputsize`?

SebastianM-C:
I rebased on master and fixed formatting. I'm not sure why codecov failed.

SebastianM-C:
@avik-pal Looks like CI passes, the codecov part at the end of the tests is unrelated and seems to randomly fail.
Is there anything left to address here?

PR #503, opened 2024-02-26
Add `outputsize` for `Chain`

SebastianM-C:
For `PairwiseFusion`, `Chain` & `Maxout`, I think I have the correct definition, but I'm not sure how to approach `Parallel` and `SkipConnection`.

Should we add `x` as an argument to `outputsize` and call `connection(l(x) for l in layers)`?

I was thinking that if we have nesting we need to define the `outputsize` recursively, so I was wondering if it would make sense to have a way to define how the connection operates on outputsizes, and maybe have dispatched for more common functions like `vcat` and `+`.

Edit: I think it would be better to first add just the simple cases in to unblock https://github.com/JuliaSymbolics/Symbolics.jl/pull/1054 and then the more complicated cases can be added in a separate PR.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Lux"
 uuid = "b2108857-7c20-44ae-9111-449ecde12c47"
 authors = ["Avik Pal <avikpal@mit.edu> and contributors"]
-version = "0.5.18"
+version = "0.5.19"
 
 [deps]
 ADTypes = "47edcb42-4c32-4615-8424-f2b9edc5f35b"
@@ -66,7 +66,7 @@ LinearAlgebra = "1.9"
 Logging = "1.9"
 LuxAMDGPU = "0.2.2"
 LuxCUDA = "0.3.2"
-LuxCore = "0.1.8"
+LuxCore = "0.1.11"
 LuxDeviceUtils = "0.1.14"
 LuxLib = "0.3.10"
 LuxTestUtils = "0.1.15"
--- a/src/layers/containers.jl
+++ b/src/layers/containers.jl
@@ -500,6 +500,8 @@ Base.length(c::Chain) = length(c.layers)
 Base.lastindex(c::Chain) = lastindex(c.layers)
 Base.firstindex(c::Chain) = firstindex(c.layers)
 
+outputsize(c::Chain) = outputsize(c.layers[end])
+
 """
     Maxout(layers...)
     Maxout(; layers...)
--- a/test/layers/containers_tests.jl
+++ b/test/layers/containers_tests.jl
@@ -242,6 +242,7 @@ end
         x = rand(Float32, 10, 1) |> aType
         y, _ = layer(x, ps, st)
         @test size(y) == (1, 1)
+        @test Lux.outputsize(layer) == (1,)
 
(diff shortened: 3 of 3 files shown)


SebastianM-C:
@avik-pal Is it possible to define `outputsize` directly from the layer definition or should we have something like `outputsize(layer, x)`? I'm not sure it's possible to define the output size without calling the `connection`s

Also, I think that my definition for `PairwiseFusion` is not correct, so I'll just remove that.

avik-pal:
We will have to do `outputsize(layer, x)`. Will it work for Symbolics if the output size is not static?

avik-pal (approved):

SebastianM-C:
Should I make a PR now with changing all `outputsize`s to 2 args to avoid a breaking release?

avik-pal:
If `outputsize(layer, x)` is fine, we can add a generic fallback in `LuxCore` to do

```julia
function outputsize(layer::AbstractExplicitLayer, x)
    ps, st = Lux.setup(rng, layer)
    y = first(layer(x, ps, st))
    return <determine the size here. 1 thing to note is that `y` might not be an array. We can handle the common cases via functors>
```

yes, let's make it 2 arguments. I wouldn't worry about it being breaking since no one is using it atm.

SebastianM-C:
would it make sense to have a trait that can say if we can "statically" determine the size? We could have something like

```julia
outputsize(layer::AbstractExplicitLayer, x, rng) = outputsize(has_static_size(layer), layer, x, rng)
outputsize(::Val{true}, layer, x, rng) =  outputsize(layer)

function outputsize(::Val{false}, layer, x, rng)
    ps, st = Lux.setup(rng, layer)
    y = first(layer(x, ps, st))
end
```

in LuxCore

avik-pal:
In that case, `outputsize` can simply ignore the `x`.

SebastianM-C:
https://github.com/LuxDL/LuxCore.jl/pull/21 will add that, so that we can define the more general version, but still be able to use the single argument version if needed, so that the Symbolics part is not directly coupled.

avik-pal:
I don't understand the point of `has_static_size`. To satisfy the interface now I need to define 2 things:

```julia
has_static_size(::MyLayer) = Val(true)

outputsize(::MyLayer) = <my fancy calculation>
```

Without it it is just:

```julia
outputsize(::MyLayer, x, rng::AbstractRNG) = ...
```

SebastianM-C:
the only advantage was to allow the single argument call to exist.

avik-pal:
If you really want users to just be able to define `outputsize(::Layer)` you can check via `hasmethod` which is fast on julia 1.10+

SebastianM-C:
>If you really want users to just be able to define `outputsize(::Layer)`

I don't have a strong preference.

Should I just use this PR to update everything to 3 arg form?

avik-pal:
No, keep the 1 argument; that seems like a nice convenience to have. But in that case we need the LuxCore PR to go in-first.

SebastianM-C:
Should I bump LuxCore to 0.1.11 here?

avik-pal:
yes

avik-pal (approved):

PR #558, opened 2024-03-21
Add a faster activation path

avik-pal:
* ~`sigmoid_fast` fails on GPU as pointed out in the NNlib PR. Need to work around that~
* ~Specialize `WrappedFunction` on `Broadcast.BroadcastFunction`~ --> can't do this mutation will violate purity wrt input/output arrays
- #570 pulls out the easy part of this

## Remaining Problems:

- [ ] Type Stability failure for DEQs
- [ ] NeuralPDE failure
- [ ] Baaad rebase


avik-pal:
See https://github.com/LuxDL/LuxLib.jl/pull/53 for a cleaner approach

avik-pal:
Absorbed into #591

PR #564, opened 2024-03-23
remove usage of apply in README

dlfivefifty:
The help says this:
```julia
help?> Lux.apply
  apply(model, x, ps, st)

  Simply calls model(x, ps, st)
```
In this case, surely it would be simpler to just call `model(x, ps, st)`?

This makes the change in the README to avoid others wondering "what is `apply` for?". An alternative would be to update the README or docs to explain why `apply` is actually useful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,10 +52,10 @@ ps, st = Lux.setup(rng, model) .|> device
 x = rand(rng, Float32, 128, 2) |> device
 
 # Run the model
-y, st = Lux.apply(model, x, ps, st)
+y, st = model(x, ps, st)
 
 # Gradients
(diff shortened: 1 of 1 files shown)


avik-pal:
I think I should update the docs for `apply` to explain why it should be preferred. The main reason is to allow necessary preprocessing to be done before we can forward the call to `model(x, ps, st)`. For example, if `x` is an Array of Structs for some AD backends it can lead to horrifying performance, we can work around that and such.

avik-pal:
Updated it here https://github.com/LuxDL/LuxCore.jl/pull/24

PR #626, opened 2024-05-06
Added layer for periodic inputs

nicholaskl97:
Sometimes when we're using physics-informed neural networks (PINNs), we want to solve a PDE on a periodic domain. Common practice is currently to include loss terms enforcing the periodic boundary conditions. The layer added in this PR would structurally enforce periodicity in certain input dimensions by replacing those dimensions $x$ with $\sin(2\pi x / period)$ and $\cos(2\pi x / period)$. This enforces periodicity in a smooth and exact manner at the cost of increasing the effective dimensionality of the input by the number of periodic dimensions. Hopefully, that isn't too penalizing thanks to the scaling of neural nets to higher dimensions.

### DIFF
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -537,12 +537,13 @@ end
     return vec(first(p(reshape(x, :, 1), ps, st))), st
 end
 
-@inline function (p::Periodic)(x::AbstractMatrix, ps, st::NamedTuple)
+@inline function (p::Periodic)(x::AbstractMatrix{T}, ps, st::NamedTuple) where T
+    k = convert.(T, 2π ./ p.periods)
     return (
         vcat(
             view(x, setdiff(axes(x, 1), p.dims), :),
-            sin.(2π ./ p.periods .* view(x, p.dims, :)),
-            cos.(2π ./ p.periods .* view(x, p.dims, :))
+            sin.(k .* view(x, p.dims, :)),
+            cos.(k .* view(x, p.dims, :))
         ),
         st)
 end
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -541,13 +541,13 @@ Lux.initialstates(::AbstractRNG, p::PeriodicEmbedding) = (k = 2 ./ p.periods,)
     return vec(first(p(reshape(x, :, 1), ps, st))), st
 end
 
-@inline function (p::PeriodicEmbedding)(x::AbstractMatrix{T}, ps, st::NamedTuple) where T
+@inline function (p::PeriodicEmbedding)(x::AbstractMatrix, ps, st::NamedTuple)
     other_dims = ChainRulesCore.@ignore_derivatives setdiff(axes(x, 1), p.dims)
     return (
         vcat(
             view(x, other_dims, :),
-            sinpi.(T.(st[:k]) .* view(x, p.dims, :)),
-            cospi.(T.(st[:k]) .* view(x, p.dims, :))
+            sinpi.(st[:k] .* view(x, p.dims, :)),
+            cospi.(st[:k] .* view(x, p.dims, :))
         ),
         st)
 end
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -521,7 +521,7 @@ periodicity.
 ## Inputs
 
   - `x` must be an `AbstractArray` with `issubset(idxs, axes(x, 1))`
-  - `st` must be a `NamedTuple` where `st[:k] = 2 ./ periods`, but on the same device as `x`
+  - `st` must be a `NamedTuple` where `st.k = 2 ./ periods`, but on the same device as `x`
 
 ## Returns
 
@@ -544,9 +544,9 @@ end
     other_idxs = ChainRulesCore.@ignore_derivatives setdiff(axes(x, 1), p.idxs)
     return (
         vcat(
-            view(x, other_idxs, :),
-            sinpi.(st[:k] .* view(x, p.idxs, :)),
-            cospi.(st[:k] .* view(x, p.idxs, :))
+            x[other_idxs, :],
+            sinpi.(st.k .* x[p.idxs, :]),
+            cospi.(st.k .* x[p.idxs, :])
         ),
         st)
 end
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -505,33 +505,32 @@ end
 outputsize(e::Embedding) = (e.out_dims,)
 
 """
-    PeriodicEmbedding(dims, periods)
+    PeriodicEmbedding(idxs, periods)
 
 Create an embedding periodic in some dimensions with specified periods. Dimensions not in
-`dims` are passed through unchanged, but dimensions in `dims` are moved to the end of the
+`idxs` are passed through unchanged, but dimensions in `idxs` are moved to the end of the
 output and replaced with their sines, followed by their cosines (scaled appropriately to
 have the specified periods). This smooth embedding preserves phase information and enforces
 periodicity.
 
 ## Arguments
 
-  - `dims`: Periodic dimensions
-  - `periods`: Periods of the dimensions `dims`, in the same order
+  - `idxs`: Indices of the periodic inputs
+  - `periods`: Periods of the periodic inputs, in the same order as in `idxs`
 
 ## Inputs
 
-  - `x` must be an `AbstractArray` with `issubset(dims, axes(x, 1))`
-  - `st` must be a `NamedTuple` where `st[:k] = 2 ./ periods`, but is on the same device as
-    `x`
+  - `x` must be an `AbstractArray` with `issubset(idxs, axes(x, 1))`
+  - `st` must be a `NamedTuple` where `st[:k] = 2 ./ periods`, but on the same device as `x`
 
 ## Returns
 
-  - `AbstractArray` of size `(size(x, 1) + length(dims), ...)` where `...` are the other
+  - `AbstractArray` of size `(size(x, 1) + length(idxs), ...)` where `...` are the other
     dimensions of `x`.
   - `st`, unchanged
 """
 @concrete struct PeriodicEmbedding <:AbstractExplicitLayer
-    dims
+    idxs
     periods
 end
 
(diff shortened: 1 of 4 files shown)


src/layers/basic.jl
```diff
@@ -503,3 +503,54 @@ function Base.show(io::IO, e::Embedding)
...
+
+@inline function (p::Periodic)(x::AbstractVector, ps, st::NamedTuple)
+    return vec(first(p(reshape(x, :, 1), ps, st))), st
+end
+
+@inline function (p::Periodic)(x::AbstractMatrix, ps, st::NamedTuple)
+    return (
+        vcat(
+            view(x, setdiff(axes(x, 1), p.dims), :),
+            sin.(2π ./ p.periods .* view(x, p.dims, :)),
```
avik-pal, line 544: This will typecast everything to Float64

src/layers/basic.jl
```diff
@@ -503,3 +503,54 @@ function Base.show(io::IO, e::Embedding)
...
+
+  - `x` must be an AbstractArray with `issubset(dims, axes(x, 1))`
+
+## Returns
+
+  - AbstractArray with dimensions `(size(x, 1) + length(dims), ...)` where `...` are the
+    other dimensions of `x`; the unchanged dimensions are first, then all the sines, then
+    all the cosines.
+  - Empty `NamedTuple()`
+"""
```
avik-pal, line 529: Let's add a doctest here. From the docstring it is not very obvious what is happening.

avik-pal (commented):
Create a separate category in https://github.com/LuxDL/Lux.jl/blob/main/docs/src/api/Lux/layers.md and add the layer.

Do we want to go with a different naming? Dense and Scale while common have strong precedence in other frameworks. Periodic is a very general name that would clash quite frequently.

src/layers/basic.jl
```diff
@@ -503,3 +503,55 @@ function Base.show(io::IO, e::Embedding)
...
+    periods
+end
+
+
+@inline function (p::Periodic)(x::AbstractVector, ps, st::NamedTuple)
+    return vec(first(p(reshape(x, :, 1), ps, st))), st
+end
+
+@inline function (p::Periodic)(x::A, ps, st::NamedTuple) where A <:AbstractMatrix
+    k = vec(convert(A, reshape(2π ./ p.periods, :, 1)))
```
avik-pal, line 541: there are `sinpi` and `cospi` which are more accurate that a manual transformation

src/layers/basic.jl
nicholaskl97: I believe I've fixed this now with the call to `convert`

src/layers/basic.jl
```diff
@@ -503,3 +503,56 @@ function Base.show(io::IO, e::Embedding)
...
+    periods
+end
+
+
+@inline function (p::Periodic)(x::AbstractVector, ps, st::NamedTuple)
+    return vec(first(p(reshape(x, :, 1), ps, st))), st
+end
+
+@inline function (p::Periodic)(x::A, ps, st::NamedTuple) where A <:AbstractMatrix
+    k = convert(A, reshape(2 ./ p.periods, :, 1))
```
nicholaskl97, line 541: I'm seeing an error from the `@jet` test that I think is coming from this line, but I haven't used that test before and am not experienced debugging those issues. Do you see the issue? Did I get the right problem line?

avik-pal: make `dims` into a tuple and hoist the `periods` into the states. That way it is guaranteed to have the same device and such as `x`. You definitely don't want to do a `convert(A, ...)`

nicholaskl97:
> Do we want to go with a different naming? Dense and Scale while common have strong precedence in other frameworks. Periodic is a very general name that would clash quite frequently.

How about SineCosinePeriodicEmbedding? Is that too verbose? Is it more descriptive? It leaves open the possibility of other future periodic embeddings.

avik-pal:
I like `PeriodicEmbedding`, we can later add a field to the struct which can take different options like `SineCosine`, etc.

src/layers/basic.jl
nicholaskl97: Why make `dims` a tuple? The call to `view` doesn't like that. 

There's also some ambiguity in the word "dimension" here, since it can both refer to each row in `x` (e.g., the number of input dimensions of the layer) and also to `N` in `AbstractArray{T, N}` (an `AbstractMatrix` is two-dimensional regardless of how many rows and columns it has, which are the matrix's dimensions). Is there a clear way to distinguish between them? `dims` is referring to the first one. Should I replace it with "indices?"

src/layers/basic.jl
```diff
@@ -503,3 +503,61 @@ function Base.show(io::IO, e::Embedding)
...
+
+Lux.initialstates(::AbstractRNG, p::PeriodicEmbedding) = (k = 2 ./ p.periods,)
+
+@inline function (p::PeriodicEmbedding)(x::AbstractVector, ps, st::NamedTuple)
+    return vec(first(p(reshape(x, :, 1), ps, st))), st
+end
+
+@inline function (p::PeriodicEmbedding)(x::AbstractMatrix{T}, ps, st::NamedTuple) where T
+    other_dims = ChainRulesCore.@ignore_derivatives setdiff(axes(x, 1), p.dims)
+    _st = (k = T.(st[:k]), )
```
nicholaskl97, line 547: I'm not sure we actually want to have this casting here, but it seems necessary to match the type of the output and `x`. Is there a better way to approach this? Should the type that people originally used when specifying `periods` matter?

src/layers/basic.jl
```diff
@@ -503,3 +503,61 @@ function Base.show(io::IO, e::Embedding)
...
+    return (
+        vcat(
+            view(x, other_dims, :),
+            sinpi.(_st[:k].* view(x, p.dims, :)),
+            cospi.(_st[:k] .* view(x, p.dims, :))
+        ),
+        _st)
```
avik-pal, line 554: I would not return the type-casted `st`. changing type of `st` is permitted but makes life much harder when used with StatefulLuxLayer

src/layers/basic.jl
avik-pal: Generally, I prefer to just use anything that is supplied and let julia semantics determine what happens. The main issue with these sort of type-casting (except in situations where without casting you hit an error) is we don't have a way to turn them off from outside.

nicholaskl97: Okay, I've removed the type-casting. I was mostly trying to address your earlier comment about accidentally type-casting everything to Float64.

avik-pal: Ah right, I was referring to using $\pi$ directly which forces a conversion, here users are kind of asking for it but supplying a different eltype in the states

src/layers/basic.jl
nicholaskl97: So, even without the call to `convert`, I'm still getting an error from that `@jet` and `@test_gradients` tests that I don't know how to diagnose or resolve. Any tips? Do you see what's going wrong?

avik-pal: Run `@report_call` and `report_opt` from JET, it will tell you where the problem is coming from.

For `@test_gradients` can you show the failure? It should print which backend is not matching against Zygote

nicholaskl97: Looking through the `@reportcall` output, and doing a bit of debugging, it looks like (perhaps among other issues) `view(x, any_idxs, :)` or similarly `x[any_idxs, :]` causes some kind of type-inference problem on the GPU according to JET. I'm not really sure where to go from there, since we need to be able to select only some rows of the input since we want to be able to create networks that are periodic in only some of the inputs. 

It's possible that the type issues are actually what's causing the error when it tries running the `@test_gradients` test, since the error is a MethodError about an ambiguous function call (two different versions of `_unsafe_setindex!`) when Tracker is calculating the gradient. 

Is there some other way I'm supposed to grab some rows, but not others? The layer is supposed to do something like `[x, y, z] -> [x, z, sin(y), cos(y)]`, so not being able to select a subset of the rows is basically a non-starter.

I appreciate the help, as I'm not very well-versed on the whole multiple-dispatch/type-stability stuff, especially when you mix in the GPU.

avik-pal: I see, if the inference problem is purely downstream, i.e. if you set `target_modules` to be Lux,LuxLib,LuxCore then it shouldn't really be a concern. It does seem test_opt fails even with target modules set https://buildkite.com/julialang/lux-dot-jl/builds/1930#018f59fb-fdbf-4b03-8902-985fd77bca10/333-1009.

I don't like this solution but this would work:

```julia
@inline __maybe_view(x, args...) = view(x, args...)  # put in utils.jl
@inline Lux.__maybe_view(x::TrackedArray, args...) = x[args...]  # put in the TrackerExt file
```

then instead of view use `__maybe_view`, this won't be optimal for Tracker but that is fine

src/layers/basic.jl
```diff
@@ -503,3 +503,58 @@ function Base.show(io::IO, e::Embedding)
...
+    return vec(first(p(reshape(x, :, 1), ps, st))), st
+end
+
+@inline function (p::PeriodicEmbedding)(x::AbstractMatrix, ps, st::NamedTuple)
+    other_idxs = ChainRulesCore.@ignore_derivatives setdiff(axes(x, 1), p.idxs)
+    return (
+        vcat(
+            view(x, other_idxs, :),
+            sinpi.(st[:k] .* view(x, p.idxs, :)),
+            cospi.(st[:k] .* view(x, p.idxs, :))
```
avik-pal, line 549:
```suggestion
            sinpi.(st.k .* view(x, p.idxs, :)),
            cospi.(st.k .* view(x, p.idxs, :))
```

avik-pal (commented):
just for style consistency

test/layers/basic_tests.jl
```diff
@@ -87,7 +87,7 @@
                     atol = 5 * eps(Float32)
                 ))
 
-            @jet layer(x, ps, st)
+            @jet layer(x, ps, st) target_modules = (Lux, LuxCore, LuxLib)
```
avik-pal, line 90: No you don't need to specify this here, `@jet` uses the Preferences file to pick this up. But for testing locally you will have to specify them

nicholaskl97: Gotcha. I'll revert that commit.

src/layers/basic.jl
nicholaskl97: It looks like removing calls to `view` solved all the JET and gradient problems. Any way I tried having a view caused some kind of type issue with CUDA. I know it's not ideal, but hopefully it won't impact performance too much.

src/layers/basic.jl
nicholaskl97: I've updated the docstring and added a doctest. Please let me know if it's clear now. I think that was also the last edit needed for all the tests to pass, so I believe it's ready to review. Thanks for all your help with this.

avik-pal (approved):

PR #640, opened 2024-05-12
Enzyme Testing + Caching in `compute_gradients`

avik-pal:
- [x] Add testing for normalization functions for Enzyme
- [x] Caching in Training
    - [x] Inplace update for Optimisers
    - [x] Enzyme Training Utilities
- [x] Rewrite the caching for Enzyme
- [x] Needs some tests
    - [x] Train a simple enough MLP in the tests


avik-pal:
We should be caching the parameter gradients loss function compiled trace and such but this should be good initial version, we anyways need a redesign of the training API later on.

ext/LuxEnzymeExt.jl
```diff
@@ -0,0 +1,26 @@
...
+        Enzyme.ReverseSplitWithPrimal, Enzyme.Const{typeof(objective_function)},
+        Enzyme.Active, Enzyme.Const{typeof(ts.model)},
+        Enzyme.Duplicated{typeof(ts.parameters)},
+        Enzyme.Const{typeof(ts.states)}, Enzyme.Const{typeof(data)})
+    tape, (loss, st_new, stats), shadow_result = fwd(
+        Enzyme.Const(objective_function), Enzyme.Const(ts.model),
+        Enzyme.Duplicated(ts.parameters, dps), Enzyme.Const(ts.states), Enzyme.Const(data))
+    rev(Enzyme.Const(objective_function), Enzyme.Const(ts.model),
+        Enzyme.Duplicated(ts.parameters, dps), Enzyme.Const(ts.states), Enzyme.Const(data),
+        (one(loss), Enzyme.make_zero(st_new), Enzyme.make_zero(stats)), tape)
```
wsmoses, line 21: If you're just computing the derivatives directly here it is more efficient to just use autodiff rather than autodiff_thunk (which may be slower for requiring things to be preserved from fwd pass to a later rev pass and assumes data may be overwritten from fwd to rev)

wsmoses: It also may be cool to extend the API to now for dps to be re-allocated every training step. You can re-use the same dps allocation for the gradient from iteration to iteration (as long as it's zero'd in place before each gradient call)

avik-pal: > If you're just computing the derivatives directly here it is more efficient to just use autodiff rather than autodiff_thunk (which may be slower for requiring things to be preserved from fwd pass to a later rev pass and assumes data may be overwritten from fwd to rev)

the objective function returns `loss, st_new, stats`, so I need to seed the st_new and stats to have no derivative, how would I do that using the `autodiff` API?

avik-pal: > It also may be cool to extend the API to now for dps to be re-allocated every training step. You can re-use the same dps allocation for the gradient from iteration to iteration (as long as it's zero'd in place before each gradient call)

Yes, I agree. We can probably cache the forward and reverse functions as well since they should stay constant.

I want to do that together with https://github.com/LuxDL/Lux.jl/issues/544

avik-pal: > Yes, I agree. We can probably cache the forward and reverse functions as well since they should stay constant.

Actually, I think I can do it in this PR itself without any breaking change

avik-pal:
Need to wait for SciMLSensitivity https://github.com/SciML/SciMLSensitivity.jl/pull/1046 before the doc build goes through

ext/LuxEnzymeExt.jl
avik-pal: Is there a way to specify a structural zero instead of doing it like `Enzyme.make_zero(st_new)`?

wsmoses:
Structural in what way

avik-pal:
> Structural in what way

As in I want to say don't backpropagate wrt this value

wsmoses:
No activity annotations (eg to differentiate or not to differentiate) are
presently at an argument or return level

avik-pal:
So how would I annotate the return type? I am getting a tuple containing a scalar, named tuple and an arbitrary object, we don't need to backpropagate for the last two

wsmoses:
Honestly I would just pass in a function which first calls the first
function

avik-pal:
you mean something like 

```julia
function compute_gradients(........)
	st_new_outer = Ref()
    stats_outer = Ref()

    function wrapper_function(args...)
	    y, st_new, stats = objective_function(args...)
	    st_new_outer[] = st_new
	    stats_outer[] = stats
	    return y
	end

    .....
end
```

wsmoses:
Yeah

ext/LuxEnzymeExt.jl
avik-pal: This is now done, first iteration creates the caches and we reuse it in subsequent iterations

avik-pal:
```julia

using ADTypes, Lux, Random, Enzyme, Optimisers

model = Chain(Conv((3, 3), 3 => 6), GroupNorm(6, 3, gelu), Conv((3, 3), 6 => 32),
    BatchNorm(32, gelu), GlobalMeanPool(), FlattenLayer(), Dense(32, 1))

x = rand(Float32, 32, 32, 3, 4);
tstate = Lux.Experimental.TrainState(Xoshiro(0), model, Adam(0.001f0));

function obj_fn(model, ps, st, x)
    y, st_new = model(x, ps, st)
    return sum(abs2, y), st_new, (;)
end

grads, loss, stats, tstate_new = Lux.Experimental.compute_gradients(
    AutoEnzyme(), obj_fn, x, tstate);

grads, loss, stats, tstate_new = Lux.Experimental.compute_gradients(
    AutoEnzyme(), obj_fn, x, tstate_new);

@btime Lux.Experimental.compute_gradients($AutoEnzyme(), $obj_fn, $x, $tstate);
# 14.726 ms (461 allocations: 9.75 MiB)

@btime Lux.Experimental.compute_gradients($AutoEnzyme(), $obj_fn, $x, $tstate_new);
# 14.233 ms (447 allocations: 9.74 MiB)
```

ext/LuxEnzymeExt.jl
```diff
@@ -0,0 +1,90 @@
...
+    ts_new = __construct_new_trainstate(
+        st_new, ts.states, forward, reverse, ts, objective_function, dps)
+    return dps, loss, stats, ts_new
+end
+
+function __compute_gradients!(
+        forward::F, reverse::R, obj_fn::O, model, ps, dps, st, data) where {F, R, O}
+    pps = Duplicated(ps, dps)
+    args = (Const(obj_fn), Const(model), pps, Const(st), Const(data))
+    tape, (loss, st_new, stats), shadow_result = forward(args...)
```
wsmoses, line 65: bumping suggestion here to merge into single autodiff call, otherwise Enzyme will conservatively assume data between forward and reverse is changed and cache a lot more [e.g. like the entire model]. There are flags you can add to say what isn't overwritten, but that's needless here.

avik-pal: Done

avik-pal:
Ok I did something wrong, it segfaulted the training test https://github.com/LuxDL/Lux.jl/actions/runs/9056705562/job/24879628489?pr=640#step:6:739

avik-pal:
Locally things pass. Now we need to wait for SciMLSensitivity compats to be updated.

avik-pal:
CI is not picking up on the latest SciMLSensitivity

PR #643, opened 2024-05-13
Add Enzyme to benchmark infra

avik-pal:
https://github.com/LuxDL/Lux.jl/pull/641 github actions seems to unable to post comment or push, so checking if that is a permissions issue

### DIFF
--- a/bench/helpers.jl
+++ b/bench/helpers.jl
@@ -72,11 +72,6 @@ function __benchmark_reverse_pass(
     return
 end
 function __benchmark_reverse_pass(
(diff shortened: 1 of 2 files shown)


avik-pal:
@wsmoses seems like Enzyme is doing well for the smaller models, but not so much for the larger ones. I would guess it is because of the broadcasting issue because most of the Lux models do heavily rely on broadcasting.

wsmoses:
Yeah we should be able to improve the broadcasting perf. For now, let's add to the continuous benchmarking to keep track of over time, and open relevant perf issues on Enzyme.

PR #659, opened 2024-05-19
Restore the rrule for merge

avik-pal:
@chrisrackauckas this will fix the SciMLSensitivity error. Not sure what is the underlying issue though, this is clearly a type piracy which shouldn't be needed.


ChrisRackauckas:
@oxinabox can we upstream this then?

oxinabox:
Yes we can. I will move it, and add tests.

PR #668, opened 2024-05-28
Add CartesianEmbedding layer

ldeso:
Lux currently offers an [Embedding](https://lux.csail.mit.edu/v0.5.51/api/Lux/layers#Lux.Embedding) layer that [calls](https://github.com/LuxDL/Lux.jl/blob/v0.5.51/src/layers/basic.jl#L500) the [NNlib.gather](https://fluxml.ai/Flux.jl/v0.14.15/models/nnlib/#NNlib.gather) method with [linear indices as arguments](https://github.com/FluxML/NNlib.jl/blob/v0.9.17/src/gather.jl#L41).

This pull request adds a CartesianEmbedding layer that allows using the other NNlib.gather method with [Cartesian indices as arguments](https://github.com/FluxML/NNlib.jl/blob/v0.9.17/src/gather.jl#L72-L73). This eliminates the need to convert manually from linear to Cartesian indices to use NNlib.gather, which significantly improves performance and reduce the number of allocations, especially on the GPU.

**Example**

```julia
# Julia v1.10.3
using Pkg
Pkg.activate(; temp=true)
Pkg.add(; url="https://github.com/ldeso/Lux.jl", rev="add-cartesianembedding-layer")
using Lux, Random

weights = [10.0 20.0 30.0
           40.0 50.0 60.0]
i = [1, 1, 1, 2, 2]
j = [1, 2, 3, 1, 2]
idx = (i, j)

rng = Random.default_rng()
model = CartesianEmbedding(size(weights) => 1; init_weight=Returns(weights))

ps, st = Lux.setup(rng, model)

model(idx, ps, st)[1]  # returns [10.0, 20.0, 30.0, 40.0, 50.0]
```

Closes #670

### DIFF
--- a/docs/src/api/Lux/layers.md
+++ b/docs/src/api/Lux/layers.md
@@ -98,5 +98,6 @@ Upsample
 ## SciML Layers
 
 ```@docs
+CartesianEmbedding
 PeriodicEmbedding
 ```
--- a/src/Lux.jl
+++ b/src/Lux.jl
@@ -88,7 +88,7 @@ include("deprecated.jl")
 export cpu, gpu  # deprecated
 
 export Chain, Parallel, SkipConnection, PairwiseFusion, BranchLayer, Maxout, RepeatedLayer
-export Bilinear, Dense, Embedding, Scale, PeriodicEmbedding
+export Bilinear, Dense, Embedding, Scale, CartesianEmbedding, PeriodicEmbedding
 export Conv, ConvTranspose, CrossCor, MaxPool, MeanPool, GlobalMaxPool, GlobalMeanPool,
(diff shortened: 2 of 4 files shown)


ldeso:
Force-pushed to pass the method ambiguity check, also moved the tests to the "Miscellaneous Layers" test item.

avik-pal:
Does this need to be a separate layer? We can just use multiple dispatch to combine with the Embedding layer

ldeso:
I created a separate pull request with multiple dispatch to let you compare (#670). While the solution with multiple dispatch results in less code, it also requires a modification of the Embedding struct which might be a bit risky (even though it passes all tests).

avik-pal:
https://github.com/LuxDL/Lux.jl/pull/670 supersedes this

PR #670, opened 2024-05-29
Add Cartesian Embedding methods

ldeso:
This is an alternative to #668 that adds Cartesian indexing methods to the existing Embedding layer instead of creating a new CartesianEmbedding layer.

Note that this modifies the Embedding composite type from:

```julia
struct Embedding <: AbstractExplicitLayer
    in_dims::Int
    out_dims::Int
    init_weight
end
```

to:

```julia
struct Embedding{N} <: AbstractExplicitLayer
    in_dims::NTuple{N, Int}
    out_dims::Int
    init_weight
end
```

Closes #668.

### DIFF
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -492,8 +492,8 @@ This layer is often used to store word embeddings and retrieve them using indice
 end
 
 function Embedding(
-        (in_dims, out_dims)::Pair{<:Union{Integer, NTuple{N, <:Integer}}, <:Integer};
-        init_weight=randn32) where {N}
+        (in_dims, out_dims)::Pair{<:Union{Integer, NTuple{<:Any, <:Integer}}, <:Integer};
+        init_weight=randn32)
     return Embedding(in_dims, out_dims, init_weight)
 end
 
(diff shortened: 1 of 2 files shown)


ldeso:
Force-pushed because I forgot to capitalize the word "Cartesian" in the documentation.

src/layers/basic.jl
```diff
@@ -461,38 +462,45 @@ This layer is often used to store word embeddings and retrieve them using indice
...
 ## Returns
 
   - Returns the embedding corresponding to each index in the input. For an N dimensional
     input, an N + 1 dimensional output is returned.
   - Empty `NamedTuple()`
 """
-@concrete struct Embedding <: AbstractExplicitLayer
-    in_dims::Int
+@concrete struct Embedding{N} <: AbstractExplicitLayer
+    in_dims::NTuple{N, Int}
```
avik-pal, line 489: This is probably not needed, just relax the `::Int`

ldeso: Done, thanks.

ldeso: I guess it would be also possible to restrict the type of `in_dims` a little bit:

```julia
@concrete struct Embedding{T <: Union{Int, NTuple{<:Any, Int}}} <: AbstractExplicitLayer
    in_dims::T
    out_dims::Int
    init_weight
end
```

But I completely relaxed the type as you said, and let `@concrete` take care of it:

```julia
@concrete struct Embedding <: AbstractExplicitLayer
    in_dims
    out_dims::Int
    init_weight
end
```

avik-pal (approved):
Thanks for the contribution!

PR #675, opened 2024-06-02
Clean up Continuous Benchmarking

avik-pal:
## TODOs

- [ ] Write a script to consolidate the data from bson files into a dataframe or something easy to extend later on
- [ ] Write a plotting code to plot the results from `benchmark-results` and push it to `gh-pages`


avik-pal:
I will open a new PR for this

PR #698, opened 2024-06-12
Add ReverseSequence

NeroBlackstone:
I don't know if this is the correct implementation

### DIFF
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -47,6 +47,32 @@ function Base.show(io::IO, r::ReshapeLayer)
 end
 
 """
+    ReverseSequence(dims)
+
+Reverse the specified dimension `dims` of the passed array
+
+## Arguments
+
+  - `dim`: Dimension that need to be reversed.
+
+## Inputs
+
+  - `x`: AbstractArray of any shape.
+
+## Returns
+
+  - AbstractArray with the same dimensions as the input
+  - Empty `NamedTuple()`
+"""
+struct ReverseSequence <: AbstractExplicitLayer
+  dim::Int
+end
+
+@inline function (r::ReverseSequence)(x::AbstractArray, ps, st::NamedTuple)
+  return reverse(x;dims = r.dims), st
+end
+
+"""
     FlattenLayer(N = nothing)
 
 Flattens the passed array into a matrix.
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -47,13 +47,14 @@ function Base.show(io::IO, r::ReshapeLayer)
 end
 
 """
-    ReverseSequence(dims)
+    ReverseSequence(dim = nothing)
 
 Reverse the specified dimension `dims` of the passed array
 
 ## Arguments
 
-  - `dim`: Dimension that need to be reversed.
+  - `dim`: Dimension that need to be reversed. If `nothing`, for AbstractVector{T}
+	it reverses itself (dimension 1), for other arrays, reverse the dimension `ndims(x) - 1`.
 
 ## Inputs
 
@@ -64,12 +65,24 @@ Reverse the specified dimension `dims` of the passed array
   - AbstractArray with the same dimensions as the input
   - Empty `NamedTuple()`
 """
-struct ReverseSequence <: AbstractExplicitLayer
-  dim::Int
+@kwdef @concrete struct ReverseSequence <: AbstractExplicitLayer
+	dim = nothing
 end
 
-@inline function (r::ReverseSequence)(x::AbstractArray, ps, st::NamedTuple)
-  return reverse(x;dims = r.dims), st
+@inline function (r::ReverseSequence{Nothing})(x::AbstractVector{T}, ps, st::NamedTuple) where {T}
+	return (isbitstype(T) ? reverse(x) : Iterators.reverse(x)),st
+end
+
+@inline function (r::ReverseSequence{Nothing})(x::AbstractArray{T,N}, ps, st::NamedTuple) where {T, N}
+  	return reverse(x; dims = ndims(x) - 1), st
+end
+
+@inline function (f::ReverseSequence)(x::AbstractVector{T}, ps, st::NamedTuple) where {T}
+	throw(DimensionMismatch(lazy"Cannot specify dimension when input is AbstractVector{T}"))
+end
+
+@inline function (f::ReverseSequence)(x::AbstractArray{T, N}, ps, st::NamedTuple) where {T, N}
+	return reverse(x; dims = r.dim), st
 end
 
 """
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -80,11 +80,11 @@ end
 end
 
 @inline function (r::ReverseSequence)(x::AbstractVector{T}, ps, st::NamedTuple) where {T}
-	if r.dim == 1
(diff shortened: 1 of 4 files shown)


src/layers/basic.jl
```diff
@@ -30,6 +30,32 @@ function Base.show(io::IO, r::ReshapeLayer)
...
+
+  - AbstractArray with the same dimensions as the input
+  - Empty `NamedTuple()`
+"""
+struct ReverseSequence <: AbstractExplicitLayer
+  dim::Int
+end
+
+@inline function (r::ReverseSequence)(x::AbstractArray, ps, st::NamedTuple)
+  return reverse(x;dims = r.dims), st
```
avik-pal, line 56: `r.dim`.

Also formatting is off

avik-pal (changes requested):
Let's redesign the interface a bit. I would default the `dims` to `nothing`. The behavior for that would be:

1. If the input is `AbstractVector{T}` and `!isbitstype(T)` then we simply call `Iterators.reverse` (note this is lazy)
2. For bitstype AbstractVector, call `reverse(x)`.
3. For other arrays, it calls `reverse(x; dims=ndims(x) - 1)`.

This essentially prevents unnecessary copies. For eg, for Bidirectional RNN we can simply construct an iterator once and make a lazy reverse iterator using this API without copying the entire array.

If `dims` is specified,

1. `AbstractVector{T}` and `!isbitstype(T)` throws an error
2. All other array inputs call `reverse(x; dims=r.dims)`

NeroBlackstone:
> If `dims` is specified,
> 
> 1. `AbstractVector{T}` and `!isbitstype(T)` throws an error

Hi, I'm curious why `isbitstype()` is needed here. For any 1D array, we can't specify `dim` right? So why we need `isbitstype()`, it would be nice if you could give an example with specific code. Thank you very much.

NeroBlackstone:
I have updated the implementation according to your guidance, except for the doubtful points above.

NeroBlackstone:
test failed on gpu, I can't reproduce on local since I don't have AMDGPU or CUDA devices...

ludvigk:
I think it fails because `Base.Iterators.Reverse{CuArray{Float32, 1, CUDA.Mem.DeviceBuffer}}` is not `isbitstype`. CUDA doesn't like that.  Even a minimal example like this fails for that reason:
```julia
g = CUDA.zeros(1)
gr = Iterators.reverse(g)
g .* gr
```

ludvigk:
Using `reverse` instead of `Iterators.reverse` on arrays of type `GPUArraysCore.AbstractGPUArray` should fix this.

NeroBlackstone:
My bad, look like test all passed now.

Please let me know if any other changes are needed

src/layers/basic.jl
```diff
@@ -30,6 +30,45 @@ function Base.show(io::IO, r::ReshapeLayer)
...
+@inline function (r::ReverseSequence{Nothing})(x::AbstractArray{T,N}, ps, st::NamedTuple) where {T, N}
+  	return reverse(x; dims = ndims(x) - 1), st
+end
+
+@inline function (r::ReverseSequence)(x::AbstractVector{T}, ps, st::NamedTuple) where {T}
+	throw(DimensionMismatch(lazy"Cannot specify dimension when input is AbstractVector{T}"))
+end
+
+@inline function (r::ReverseSequence)(x::AbstractArray{T, N}, ps, st::NamedTuple) where {T, N}
+	return reverse(x; dims = r.dim), st
```
avik-pal, line 69: Run the formatter; tabs and spaces are inconsistent.

src/layers/basic.jl
```diff
@@ -30,6 +30,45 @@ function Base.show(io::IO, r::ReshapeLayer)
...
+@inline function (r::ReverseSequence{Nothing})(x::AbstractVector{T}, ps, st::NamedTuple) where {T}
+	return (isbitstype(T) ? reverse(x) : Iterators.reverse(x)),st
+end
+
+@inline function (r::ReverseSequence{Nothing})(x::AbstractArray{T,N}, ps, st::NamedTuple) where {T, N}
+  	return reverse(x; dims = ndims(x) - 1), st
+end
+
+@inline function (r::ReverseSequence)(x::AbstractVector{T}, ps, st::NamedTuple) where {T}
+	throw(DimensionMismatch(lazy"Cannot specify dimension when input is AbstractVector{T}"))
```
avik-pal, line 65: `dim = 1` is still valid. Error if it is not `1`

NeroBlackstone:
> Using `reverse` instead of `Iterators.reverse` on arrays of type `GPUArraysCore.AbstractGPUArray` should fix this.

I got a PC with GPU and test locally.

I found `Float64` works well but `Int` throws errors. Weird...

avik-pal:
Add this layer to the docs, also rebase with main. That will fix the doc build

src/layers/basic.jl
```diff
@@ -30,6 +30,52 @@ function Base.show(io::IO, r::ReshapeLayer)
...
+Reverse the specified dimension `dims` of the passed array
+
+## Arguments
+
+  - `dim`: Dimension that need to be reversed. If `nothing`, for AbstractVector{T}
+    it reverses itself (dimension 1), for other arrays, reverse the dimension `ndims(x) - 1`.
+
+## Inputs
+
+  - `x`: AbstractArray of any shape.
```
avik-pal, line 45:
```suggestion
  - `x`: AbstractArray.
```

src/layers/basic.jl
```diff
@@ -30,6 +30,52 @@ function Base.show(io::IO, r::ReshapeLayer)
...
+        x::AbstractArray{T, N}, ps, st::NamedTuple) where {T, N}
+    return reverse(x; dims=ndims(x) - 1), st
+end
+
+@inline function (r::ReverseSequence)(x::AbstractVector{T}, ps, st::NamedTuple) where {T}
+    if r.dim == 1
+        return reverse(x), st
+    else
+        throw(DimensionMismatch(lazy"Cannot specify a dimension other than 1 for AbstractVector{T}"))
+    end
```
avik-pal, line 71:
```suggestion
    r.dim == 1 && return reverse(x), st
    throw(DimensionMismatch(lazy"Cannot specify a dimension other than 1 for AbstractVector{T}"))
```

src/layers/basic.jl
```diff
@@ -30,6 +30,52 @@ function Base.show(io::IO, r::ReshapeLayer)
...
+  - `x`: AbstractArray of any shape.
+
+## Returns
+
+  - AbstractArray with the same dimensions as the input
+  - Empty `NamedTuple()`
+"""
+@kwdef @concrete struct ReverseSequence <: AbstractExplicitLayer
+    dim = nothing
+end
```
avik-pal, line 54:
```suggestion
@kwdef struct ReverseSequence{D <: Union{Int, Nothing}} <: AbstractExplicitLayer
    dim::D = nothing
end
```

NeroBlackstone:
I noticed some test failures, should I follow up on this PR? Or leave all the rest to you？

avik-pal:
Github won't let me edit files directly. So if you add the docs to https://github.com/LuxDL/Lux.jl/blob/main/docs/src/api/Lux/layers.md#misc-helper-layers the build will pass

avik-pal (approved):

PR #706, opened 2024-06-16
Add `ReverseSequence()` docs

NeroBlackstone:
Add `ReverseSequence()` docs

### DIFF
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -64,6 +64,24 @@ Reverse the specified dimension `dims` of the passed array
 
   - AbstractArray with the same dimensions as the input
   - Empty `NamedTuple()`
+
+## Example
+```jldoctest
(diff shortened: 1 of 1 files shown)


src/layers/basic.jl
```diff
@@ -64,6 +64,21 @@ Reverse the specified dimension `dims` of the passed array
...
+```jldoctest
+julia> model = ReverseSequence()
+ReverseSequence()
+
+julia> rng = Random.default_rng();
+       Random.seed!(rng, 0);
+       ps, st = Lux.setup(rng, model);
+       x = [1.0, 2.0, 3.0];
+
+julia> y, st_new = model(x, ps, st);
```
avik-pal, line 79:
```suggestion
julia> y, st_new = model(x, ps, st)
```

avik-pal (approved):

PR #739, opened 2024-06-28
Run CI on pre

avik-pal:


avik-pal:
@wsmoses is Enzyme currently not functional on 1.11-rc1? https://github.com/LuxDL/Lux.jl/actions/runs/9705757902/job/26788368870?pr=739#step:6:583

Issue #758, opened 2024-07-11
Feature request: TimeDistributed Layer


NeroBlackstone:
Motivation:
If [Lux.Recurrence ](https://lux.csail.mit.edu/stable/api/Lux/layers#Linear-Layers) `return_sequence` arg is `true`, it will return a vector of 2d-matrix.

``` julia
model = Chain(Recurrence(GRUCell(3=>32),return_sequence = true))

rng = Random.default_rng()
Random.seed!(rng, 0)
ps, st = Lux.setup(rng, model)

x = rand(Float32,3,4,5)
y1 = model(x,ps,st)[1]

@show size(y1)
@show size(y1[1])

# sequence_length
#size(y1) = (4,)
# output_size * batch_size
#size(y1[1]) = (32, 5)
```

Suppose we have a sequence labeling task, and the output of each time step is fed into a Dense layer (as a classifier) ​​for classification, looks like this:

![TimeDistributed](https://github.com/LuxDL/Lux.jl/assets/17470335/8d7a4a32-850a-4e12-bef5-f88abc797adb)

Can the existing layers do this? What do you think?

I think it works like this:

``` julia
model = Chain(Recurrence(GRUCell(3=>32),return_sequence = true),TimeDistributed(Dense(32=>2)))
rng = Random.default_rng()
Random.seed!(rng, 0)
ps, st = Lux.setup(rng, model)

x = rand(Float32,3,4,5)
y1 = model(x,ps,st)[1]

@show size(y1)
@show size(y1[1])

# sequence_length
#size(y1) = (4,)
# output_size * batch_size
#size(y1[1]) = (2, 5)
```

I have not studied [TimeDistributed](https://keras.io/api/layers/recurrent_layers/time_distributed/) in Keras, and I don’t know if my understanding is correct.

Is it necessary to have a function as a parameter in order to merge the output of TimeDistributed? like:

``` julia
...
TimeDistributed(Dense(32=>2);vcat)
...
# output
# sequence_length
#size(y1) = (2, 5)
```

Do you think this feature is necessary? If so, I can implement it. Please tell me some ideas (such as existing layers that can be referenced) and I will complete it. Thank you!

NeroBlackstone:
> Motivation:
> If [Lux.Recurrence ](https://lux.csail.mit.edu/stable/api/Lux/layers#Linear-Layers) `return_sequence` arg is `true`, it will return a vector of 2d-matrix.
> 
> ``` julia
[...]

Issue #426, opened 2023-10-20
Broadcast Layer


avik-pal:
A simple implementation:

```julia
struct BroadcastLayer{T <: NamedTuple} <: AbstractExplicitContainerLayer{(:layers,)}
    layers::T
end

function BroadcastLayer(layers...)
    for l in layers
        if !iszero(statelength(l))
            throw(ArgumentError("Stateful layer `$l` are not supported for `BroadcastLayer`."))
        end
    end
    names = ntuple(i -> Symbol("layer_$i"), length(layers))
    return BroadcastLayer(NamedTuple{names}(layers))
end

BroadcastLayer(; kwargs...) = BroadcastLayer(connection, (; kwargs...))

function (m::BroadcastLayer)(x, ps, st::NamedTuple{names}) where {names}
    results = (first ∘ Lux.apply).(values(m.layers), x, values(ps), values(st))
    return results, st
end

Base.keys(m::BroadcastLayer) = Base.keys(getfield(m, :layers))
```

_Originally posted by @avik-pal in https://github.com/LuxDL/Lux.jl/issues/282#issuecomment-1513524492_

NeroBlackstone:
``` julia
BroadcastLayer(; kwargs...) = BroadcastLayer(connection, (; kwargs...))
```

What is the role of `kwargs` and `connection` in this line?
Do I need to add the implementation of BroadcastLayer(connection, (; kwargs...))?

NeroBlackstone:
I still can't understand what BroadcastLayer does when there are multiple layers passed in.

1. If there are **n layers** in the BroadcastLayer, and **length of x is 1**, `x` will be copied n times and passed to every internal layers respectively.
2. If there are **n layers** in the BroadcastLayer, and **length of x is n**, `x_i` will be passed to the corresponding `layer_i`.
3. If there is only **1 layer** in the BroadcastLayer, and **length of x is n**, `x_i` will be passed to `layer` respectively.
4. If there are **n layers** in the BroadcastLayer, and **length of x is m**, m != n, m>1 and n>1, throw a error.

Is my understanding correct and is it necessary to document this?

NeroBlackstone:
@avik-pal Hi, Could you help me confirm that my understanding is correct? If its correctI will open a PR soon

avik-pal:
Honesty with the introduction of `@compact`, it might make sense not to have them inbuilt. Broadcasting is effectively:

```julia
[...]

PR #768, opened 2024-07-15
Update RNN docs

NeroBlackstone:
Related issue: #426 #758


avik-pal:
This makes more sense as a discussion question/answer rather than expanding the docstring to be larger. I know we have done so in the past for some other layers, but at some point, this becomes impossible to maintain.

In the future, we might have a "Migration Guide" from "Pytorch"/"Jax"/"Keras," but that is a much larger undertaking. Until then, it might be useful to have these in GH discussions.

PR #805, opened 2024-07-29
feat: expand nested AD support to TrackedArrays

avik-pal:
## TODOs

- [x] Initial Support
- [x] Expand the tests
- [ ] Proper support for parameter gradients
- [ ] Documentation updates


avik-pal:
Doing this on a case by case basis. see #848

PR #810, opened 2024-07-31
[WIP] SAM loss 

Vaibhavdixit02:
Adding SAM loss functionality here https://arxiv.org/abs/2010.01412

### DIFF
--- a/src/helpers/losses.jl
+++ b/src/helpers/losses.jl
@@ -629,10 +629,10 @@ function __unsafe_apply_loss(loss::SAMLoss, ŷ, y)
 end
 
 function CRC.rrule(cfg::RuleConfig{>:HasReverseMode}, SL::typeof(SAMLoss), model, ps, st, (x, y))
-    grad = Zygote.gradient(pars->SL.loss_fn(model(x, pars, st)[1], y), ps)
+    grad = CRC.rrule_via_ad(cfg, pars->SL.loss_fn(model(x, pars, st)[1], y), ps)[2]
     ϵ = SL.ρ * grad / (norm(grad) + eps)
 
-    return SL.loss_fn(model(x, ps, st)[1], y), Zygote.gradient(pars->SL.loss_fn(model(x, pars, st)[1], y), ps .+ ϵ) 
+    return SL.loss_fn(model(x, ps, st)[1], y), CRC.rrule_via_ad(pars->SL.loss_fn(model(x, pars, st)[1], y), ps .+ ϵ) 
 end
 
 
--- a/src/helpers/losses.jl
+++ b/src/helpers/losses.jl
@@ -609,6 +609,30 @@ function __unsafe_apply_loss(loss::GenericLossFunction, ŷ, y)
     return __fused_agg(loss.agg, loss.loss_fn, ŷ, y)
 end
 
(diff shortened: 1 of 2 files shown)


src/helpers/losses.jl
```diff
@@ -609,6 +609,34 @@ function __unsafe_apply_loss(loss::GenericLossFunction, ŷ, y)
...
+    ŷ, st_ = model(x, ps, st)
+    return loss.loss_fn(ŷ, y)
+end
+
+function __unsafe_apply_loss(loss::SAMLoss, ŷ, y)
+    return loss.loss_fn(ŷ, y)
+end
+
+function CRC.rrule(cfg::RuleConfig{>:HasReverseMode}, SL::typeof(SAMLoss), model, ps, st, (x, y))
+    grad = Zygote.gradient(pars->SL.loss_fn(model(x, pars, st)[1], y), ps)
```
avik-pal, line 632: Lux doesn't have a dep on `Zygote` (and can't take one). You can do this by calling `CRC.rrule_via_ad(cfg, ....)`

src/helpers/losses.jl
```diff
@@ -609,6 +609,34 @@ function __unsafe_apply_loss(loss::GenericLossFunction, ŷ, y)
...
+
+function SAMLoss(loss_fn; ρ=0.05)
+    return SAMLoss(loss_fn, ρ)
+end
+
+
+function (loss::SAMLoss)(model, ps, st, (x, y))
+    ŷ, st_ = model(x, ps, st)
+    return loss.loss_fn(ŷ, y)
+end
```
avik-pal, line 625: Is this needed? Can't it go via the size check and the usual dispatch

src/helpers/losses.jl
```diff
@@ -609,6 +609,34 @@ function __unsafe_apply_loss(loss::GenericLossFunction, ŷ, y)
...
+
+function __unsafe_apply_loss(loss::SAMLoss, ŷ, y)
+    return loss.loss_fn(ŷ, y)
+end
+
+function CRC.rrule(cfg::RuleConfig{>:HasReverseMode}, SL::typeof(SAMLoss), model, ps, st, (x, y))
+    grad = CRC.rrule_via_ad(cfg, pars->SL.loss_fn(model(x, pars, st)[1], y), ps)[2]
+    ϵ = SL.ρ * grad / (norm(grad) + eps)
+
+    return SL.loss_fn(model(x, ps, st)[1], y), CRC.rrule_via_ad(pars->SL.loss_fn(model(x, pars, st)[1], y), ps .+ ϵ)
```
avik-pal, line 635: missing `cfg`

avik-pal: `ps .+ \epsilon` won't work if `ps` is not an AbstractArray. You would want to use `recursive_map(Base.Fix1(+, eps), ps)`

Vaibhavdixit02:
@avik-pal thanks for the review! Thinking this through, it would be better to implement this as an optimization loop instead of the loss function approach I was taking here, since that'll give more control over the hyperparameters being used and also make the gradient computations use caches more effectively. I'll close this here for now and try this other approach out for a bit if I think this makes more sense I'll re-open it.

PR #897, opened 2024-09-13
docs: add an Optimization.jl tutorial showcasing lazy data movement

avik-pal:
cc @Vaibhavdixit02


Vaibhavdixit02:
Nice! Let's see what the GPU CI says

PR #910, opened 2024-09-16
Adapt to pending Enzyme breaking change

wsmoses:
Incidentally this extension doesn't seem to have any code broken by the breaking change. Nevertheless trying to open / flag early for the ecosystem

shortlist of breaking changes:

EnzymeRules for forward mode are defined differently [now with a config] ForwardMode moved DupNoNeed / Dup to specify if the return is needed or not -> Forward and ForwardWithPrimal

Runtime Activity is now in the mode

gradient/jacobian all optionall return values with the WithPrimal arg


avik-pal:
I will have to make the changes to LuxTestUtils before they get reflected in the tests

avik-pal:
#950

PR #937, opened 2024-09-18
ci(buildkite): debugging CUDA segfaults on CI

avik-pal:
trying to debug the CUDA segfaults


avik-pal:
conclusion: once 0.10 of CUDA_Driver_jll is supported (with the new CUDA release) segfaults seem to be stopped

avik-pal:
Seems like removing the compat from CUDA_Driver_jll also does the trick.

Issue #963, opened 2024-10-01
Efficient way to compute Jacobian in nested AD 


facusapienza21:
HI! I was looking at the example in the docs about how to perform [nested AD with Lux](https://lux.csail.mit.edu/stable/manual/nested_autodiff) with Lux. The code in the documentation definitively works, and I have included a full example in [Discurse](https://discourse.julialang.org/t/nested-and-different-ad-methods-altogether-how-to-add-ad-calculations-inside-my-loss-function-when-using-neural-differential-equations/108985/10) with this for completeness. However, I noticed that when we evaluate the Jacobian this give us the Jacobian full with zeros, given than outputs of different inputs are not algebraically related (that is, no need to compute the full matrix). 

For the following piece of code, 
```julia
function loss(model, ps, st)

    # Compute predicions using model parameters
    X_pred = predict(ps)
    loss_emp = mean(abs2, Xₙ .- X_pred)

    # Make it a stateful layer
    smodel = StatefulLuxLayer{true}(U, ps, st)
    
    J = ForwardDiff.jacobian(smodel, Xₙ)
    loss_reg = 0.01f0 * abs2(norm(J))
    return loss_emp + loss_reg
end
```
this is how `J` looks like:
![image](https://github.com/user-attachments/assets/34e245ad-f329-4ef5-93b9-87204d17cf5b)

I think this is not very efficient, but I am also maybe missing something of how Lux internally manages this calculations. I tried computing the Jacobian/gradient for each individual input layer value, and this seems to be very inefficient. Other options on the top of my head include
- Computing a VJP of this `J` times a vector with ones instead, to avoid the calculation of the zero entries. I think this is what [Lux.jacobian_vector_preoduct](https://lux.csail.mit.edu/stable/api/Lux/autodiff#Nested-2nd-Order-AD) is doing?
- Add some sparsity pattern to the Jacobian calculation 

Any suggestion here? I just would like to see this example using the best practices when using Lux. 

Thank you so much! All this looks amazing.

avik-pal:
Use `batched_jacobian` https://lux.csail.mit.edu/stable/api/Lux/autodiff#Lux.batched_jacobian

avik-pal:
Only thing to be aware of is that it gives you a 3D array (essentially a BlockDiagonal Matrix without storign the zeros)

facusapienza21:
Lovely, just like this I guess?
[...]

PR #964, opened 2024-10-01
Added to Nested AD example how to use `batched_jacobian`

facusapienza21:
Just following on the discussion with @avik-pal in #963. 

I am not sure about the formatting of this, I hope there is some CI in place to be sure all looks nice. Happy to make any changes @avik-pal that you think adequate.

### DIFF
--- a/docs/src/manual/nested_autodiff.md
+++ b/docs/src/manual/nested_autodiff.md
@@ -125,7 +125,7 @@ function loss_function_batched(model, x, ps, st, y)
     smodel = StatefulLuxLayer{true}(model, ps, st)
     ŷ = smodel(x)
     loss_emp = sum(abs2, ŷ .- y)
-    # You can use `Zygote.jacobian` as well but ForwardDiff tends to be more efficient here
+    # You can use `AutoZygote()` as well but `AutoForwardDiff()` tends to be more efficient here
     J = batched_jacobian(smodel, AutoForwardDiff(), x)
     loss_reg = abs2(norm(J .* 0.01f0))
     return loss_emp + loss_reg
--- a/docs/src/manual/nested_autodiff.md
+++ b/docs/src/manual/nested_autodiff.md
@@ -106,15 +106,52 @@ nothing; # hide
 That's pretty good, of course you will have some error from the finite differences
 calculation.
 
-!!! tip
+### Using Batched Jacobian for Multiple Inputs 
 
-    Notice that in this example the Jacobian `J` consists on the full matrix of derivatives of `smodel` with respect
-    the different inputs in `x`. In order to avoid the unnecessary calculation of zero entries of the Jacobian, you can 
-    instead use [`batched_jacobian`](@ref) to parse the calculation of the Jacobian per each single input. 
-    ```@example nested_ad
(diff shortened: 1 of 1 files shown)


docs/src/manual/nested_autodiff.md
```diff
@@ -106,6 +106,16 @@ nothing; # hide
 That's pretty good, of course you will have some error from the finite differences
 calculation.
 
+!!! tip
+
+    Notice that in this example the Jacobian `J` consists on the full matrix of derivatives of `smodel` with respect
+    the different inputs in `x`. In order to avoid the unnecessary calculation of zero entries of the Jacobian, you can 
+    instead use [`batched_jacobian`](@ref) to parse the calculation of the Jacobian per each single input. 
+    ```@example nested_ad
+    J = batched_jacobian(smodel, AutoForwardDiff(), x)
```
avik-pal, line 115: make it into a full function like `loss_function1`. Then if you add the checks with `@assert ... # hide` those will ensure that the computation is correct and won't be exposed in the docs

facusapienza21: I will do it! But please take a look at my following comment.

facusapienza21:
@avik-pal the current example in the documentation includes a neural network with batch normalization, reason why the Jacobian has non-trivial entries everywhere. This is quite different than the original example I posted in Discourse, where the idea is to compute the gradient with respect to the input layer multiple times. Here we can do two things: 
- Remove the Batch normalization of the first example to reproduce that behaviour, but I guess you include it for good reasons
- Make this new example different and remove the batch normalization of the chain. 

Maybe the second option makes more sense, but in that case I will add more explanation in the text. What do you think? The two options look fine for me. 

For reference, here is `J` with BatchNorm 
image
and without it
image

avik-pal:
Ah right forgot about that. You can change that to `GroupNorm`. I put in a normalization layer to ensure that we aren't cheating with Zygote over Zygote (which will fail for any of the normalization layers)

avik-pal:
Alternatively you can add the `smodel = ...` in that example block. currently it doesn't have it in global scope https://buildkite.com/julialang/lux-dot-jl/builds/4386#019245db-55f8-484c-ac71-b18b7a56b970/656-1623

facusapienza21:
Thank you @avik-pal , I just made this changes, let me know what do you think.

docs/src/manual/nested_autodiff.md
```diff
@@ -106,6 +106,53 @@ nothing; # hide
...
+ps, st = Lux.setup(StableRNG(0), model)
+x = randn(StableRNG(0), Float32, 2, 10)
+y = randn(StableRNG(11), Float32, 2, 10)
+
+function loss_function_batched(model, x, ps, st, y)
+    # Make it a stateful layer
+    smodel = StatefulLuxLayer{true}(model, ps, st)
+    ŷ = smodel(x)
+    loss_emp = sum(abs2, ŷ .- y)
+    # You can use `Zygote.jacobian` as well but ForwardDiff tends to be more efficient here
```
avik-pal, line 128:
```suggestion
    # You can use `AutoZygote()` as well but `AutoForwardDiff()` tends to be more efficient here
```

avik-pal (approved):
Thanks!

PR #965, opened 2024-10-01
Remove line about "not saving the model"

asinghvi17:
Not sure what this is but it seems counterintuitive (or incomplete).  Feel free to reject or modify.

### DIFF
--- a/examples/SimpleRNN/main.jl
+++ b/examples/SimpleRNN/main.jl
@@ -173,8 +173,7 @@ nothing #hide
 # ## Saving the Model
 
 # We can save the model using JLD2 (and any other serialization library of your choice)
-# Note that we transfer the model to CPU before saving. Additionally, we recommend that
-# you don't save the model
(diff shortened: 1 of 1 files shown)


examples/SimpleRNN/main.jl
```diff
@@ -173,8 +173,7 @@ nothing #hide
 # ## Saving the Model
 
 # We can save the model using JLD2 (and any other serialization library of your choice)
-# Note that we transfer the model to CPU before saving. Additionally, we recommend that
-# you don't save the model
+# Note that we transfer the model to CPU before saving.
```
avik-pal, line 176:
```suggestion
# Note that we transfer the model to CPU before saving. Additionally, we recommend that
# you don't save the model struct and only save the parameters and states.
```

How about this?

asinghvi17: sounds good, thanks!

asinghvi17:
Spell check failure seems unrelated?

PR #967, opened 2024-10-02
docs: add compilation example using Reactant

avik-pal:


docs/src/manual/compiling_lux_models.md
```diff
@@ -0,0 +1,171 @@
...
+
+∂ps_zyg = only(Zygote.gradient(ps -> loss_function(model, ps, st, x, y), ps))
+```
+
+Now we will compile the gradient function using `Reactant.@compile`.
+
+```@example compile_lux_model
+function enzyme_gradient(model, ps, st, x, y)
+    dps = Enzyme.make_zero(ps)
+    Enzyme.autodiff(Enzyme.Reverse, Const(loss_function), Active, Const(model),
```
wsmoses, line 106: if dps isn't preallocated there's now some nicer syntax of 

```julia
Enzyme.gradient(Reverse, Const(loss_function), Const(model), ps, Const(st), Const(x), Const(y))
```

avik-pal: BTW TPUs also working nicely https://colab.research.google.com/drive/1kVP_wLI4vFhR32ggImdssCj0xI7XR00u?usp=sharing

PR #673, opened 2024-06-01
Compile training loop with Reactant

avik-pal:
Pulling out training logic from #665, because this is simpler to implement and merge.

## Example Usage

We can simply replace `AutoEnzyme` or any of those backends with `AutoReactant` and use `single_train_step` or `single_train_step!` and every

TODO: Code example

## Upstream Needs

- [x] https://github.com/EnzymeAD/Reactant.jl/issues/14
- [x] https://github.com/EnzymeAD/Reactant.jl/pull/34

## TODOs

- [x] Fallback implementation for existing backends
- [ ] Cleanup the code
- [ ] Updates to `ReactantBackend`
  - [ ] Add option to force return gradients (warn users that this is going to slow down their code) -- We can use caching to probably reduce the penalty for this, but I will do that later.
- [ ] Reactant Backend
  - [x] `single_train_step!`
  - [ ] single_train_step
- [ ] Common Optimisers
  - [x] Descent
  - [ ] Adam
- [ ] Add to documentation
  - [ ] Highlight how we handle optimisers. Shared parameters won't work for the time being. Once we use full fledged Optimisers.jl that will resume working as usual.
  - [ ] Let's create a manual page listing current shortcomings and deviations from the other backends -- 1 example is we don't return the gradient, instead `
- [ ] Testing
  - [ ] Common model compile testing
  - [ ] inference testing

PR #969, opened 2024-10-04
feat: compile training loop automatically using reactant

avik-pal:
Much less intrusive than https://github.com/LuxDL/Lux.jl/pull/673. We automatically check for `get_device_type` and use Reactant if paired with Enzyme.

## TODOs

- [x] Reactant Backend -- not an actual backend but trigger with a combination of `xla_device` and `AutoEnzyme`
  - [x] `single_train_step!`
  - [x] `single_train_step`
  - [x] `compute_gradients`
- [x] Optimisers are not compiling. We have 2 solutions here
  - ~We hack around a proxy for the optimizer similar to what I was doing in #673.~
  - Actually stare at the error and fix it in Reactant proper (I kind of prefer going down this route) -- https://github.com/EnzymeAD/Reactant.jl/pull/156
- [x] Upstream
  - [x] https://github.com/EnzymeAD/Reactant.jl/pull/161
- [x] Testing
  - [x] Common model compile testing
  - [x] loss function compilation
  - [x] re-enable all tests

## Future PRs

- Docs will be added in #967 
- Inference? What would be a good API for that


avik-pal:
## Example Usage

```julia
using Lux, Reactant, Enzyme, Random
using Optimisers

dev = xla_device()

model = Chain(
    Dense(2 => 32, tanh),
    Dense(32 => 32, tanh),
    Dense(32 => 2)
)
ps, st = Lux.setup(Random.default_rng(), model) |> dev

x = randn(Float32, 2, 1024) |> dev
y = x .^ 2 .+ 1

function sse(model, ps, st, (x, y))
    z, stₙ = model(x, ps, st)
    diff = z .- y
    return sum(abs2, diff), stₙ, (;)
end

train_state = Training.TrainState(model, ps, st, Descent(0.01))

grads, loss, stats, train_state_compiled = Training.single_train_step!(
    AutoEnzyme(),
    sse,
    (x, y),
    train_state
)

grads, loss, stats, train_state_compiled = Training.single_train_step!(
    AutoEnzyme(),
    sse,
    (x, y),
    train_state_compiled
)
```

Using any other ADType will cause an error.

avik-pal:
https://github.com/EnzymeAD/Reactant.jl/pull/161 is needed for the loss functions tests to pass

PR #986, opened 2024-10-18
fix: `LV`/`Octavian` moved to optional deps

avik-pal:
xref https://github.com/LuxDL/LuxLib.jl/pull/175


avik-pal:
docs fixed by https://github.com/LuxDL/MLDataDevices.jl/pull/87

PR #987, opened 2024-10-23
docs(reactant): simplify the enzyme call

avik-pal:


docs/src/manual/compiling_lux_models.md
```diff
@@ -102,10 +102,8 @@ Now we will compile the gradient function using `Reactant.@compile`.
 
 ```@example compile_lux_model
 function enzyme_gradient(model, ps, st, x, y)
-    dps = Enzyme.make_zero(ps)
-    Enzyme.autodiff(Enzyme.Reverse, Const(loss_function), Active, Const(model),
-        Duplicated(ps, dps), Const(st), Const(x), Const(y))
-    return dps
+    return only(Enzyme.autodiff(Enzyme.Reverse, Const(loss_function), Active, Const(model),
```
wsmoses, line 105: oh this should be Enzyme.gradient if you're using it this way, not enzyme.autodiff

docs/src/manual/compiling_lux_models.md
```diff
@@ -102,10 +102,8 @@ Now we will compile the gradient function using `Reactant.@compile`.
...
+        ps, Const(st), Const(x), Const(y)))
```
avik-pal, line 106:
```suggestion
    return Enzyme.gradient(Enzyme.Reverse, Const(loss_function), Const(model),
        ps, Const(st), Const(x), Const(y))[2]
```

PR #996, opened 2024-10-29
Move code blocks inside bullet list

abhro:
In gpu_management.md, fix list fragmentation by putting each code block as part of its list item.


avik-pal (changes requested):
This doesn't seem to work https://buildkite.com/julialang/lux-dot-jl/builds/4535#0192d952-60b6-4938-bc1e-358dee4371c0/763-1645

avik-pal (approved):

PR #1020, opened 2024-11-03
feat: support passing in device and client to XLA

avik-pal:


lib/MLDataDevices/src/public.jl
```diff
@@ -9,8 +9,10 @@ end
 struct MetalDevice <: AbstractGPUDevice end
 struct oneAPIDevice <: AbstractGPUDevice end
 
-# TODO: Later we might want to add the client field here?
-struct XLADevice <: AbstractAcceleratorDevice end
+@kwdef struct XLADevice{C, D} <: AbstractAcceleratorDevice
```
wsmoses, line 12: For future extensibility [as we get distributed and other things setup], can this be renamed to ReactantDevice?

avik-pal: yeah I can do it with a `const XLADevice = ReactantDevice` to preserve backward compatibility.

wsmoses: Sounds good!

avik-pal:
I want to wait for https://github.com/EnzymeAD/Reactant.jl/pull/222 which exposes `device` so that we don't have to make direct XLA calls

PR #1026, opened 2024-11-04
fix: init hidden state for reactant

### LINKED ISSUE #1025: Compiling Recurrent Models with Reactant
```julia
using Lux, Reactant, Random

model = Recurrence(RNNCell(4 => 4))
ps, st = Lux.setup(Xoshiro(123), model) |> Reactant.to_rarray
x = rand(Float32, 4, 16, 12) |> Reactant.ConcreteRArray

@code_hlo model(x, ps, st)
```

The issue originates from the following function moving the data to a ConcreteRArray instead of a TracedRArray when run inside a compilation context. I could use `@reactant_override` to define custom dispatches when used inside `Reactant.compile`, but I am not sure that should be the recommended usage

```julia
function init_rnn_hidden_state(rng::AbstractRNG, rnn, x::AbstractMatrix)
    # TODO: Once we support moving `rng` to the device, we can directly initialize on the
    #       device
    return rnn.init_state(rng, rnn.out_dims, Base.size(x, 2)) |> get_device(x)
end
```

avik-pal:
fixes #1025


avik-pal:
Once we tag a new Reactant release I can add tests and merge this

avik-pal:
only waiting for a new Reactant release. Then I will bump the minimum version requirements

avik-pal:
need to fix some of the scalar indexing issues now that scalar indexing is disabled in Reactant for non-CPU buffers

PR #1039, opened 2024-11-05
Always use fmap and bump functors to 0.5

### LINKED ISSUE #1017: sending to devices tuples, named tuples and arrays does not keep track of identical objects
The issue has been reported in 
https://github.com/FluxML/Flux.jl/issues/2513
and is due to these specializations
https://github.com/LuxDL/MLDataDevices.jl/blob/17419d27888e3a48b52f318249a0f037524f0f1e/src/public.jl#L341
The simplest solution would be to remove the specializations and let `fmap` handle everything.

Possible optimized implementations could be upstreamed to Functors.jl, but we really want to keep track of object identity here.

CarloLucibello:
Fix #1017 
Always use `fmap` for traversing types, removing tuple/namedtuple/array specialization and allowing for identity to be preserved in data transfer.


CarloLucibello:
testing needs https://github.com/jonniedie/ComponentArrays.jl/pull/273 to be merged

avik-pal:
absorbing this into https://github.com/LuxDL/Lux.jl/pull/1069 since it needed a few more changes. Anticipating to finish it by this Friday.

As an aside the generated function update for type stability is quite some brilliant work!

PR #1041, opened 2024-11-05
fix: make enzyme testing opt-in for now

avik-pal:


wsmoses:
so as of last night 1.11 support is essentially in, can we test and sees if it resolves?

avik-pal:
Ah nice, I will trigger it

PR #1042, opened 2024-11-06
test: try re-enabling enzyme testing on 0.13.16

avik-pal:


avik-pal:
Need to also reenable some of the tests manually in LuxLib

avik-pal:
@wsmoses any idea what the following means:

```julia
initfn=define private void @jlplt_ijl_set_task_tid_448994({} addrspace(10)* %0, i32 %1) #9 {
top:
  %2 = load atomic void ()*, void ()** null unordered, align 8
  %3 = icmp ne void ()* %2, null
  br i1 %3, label %ccall, label %dlsym

dlsym:                                            ; preds = %top
  %4 = call void ()* @ijl_load_and_lookup(i8* inttoptr (i64 3 to i8*), i8* getelementptr inbounds ([17 x i8], [17 x i8]* @_j_str_ijl_set_task_tid_21, i32 0, i32 0), i8** @jl_libjulia_internal_handle)
  store atomic void ()* %4, void ()** null release, align 8
  br label %ccall

ccall:                                            ; preds = %dlsym, %top
  %5 = phi void ()* [ %2, %top ], [ %4, %dlsym ]
  %6 = bitcast void ()* %5 to void ({} addrspace(10)*, i32)*
  %7 = bitcast void ({} addrspace(10)*, i32)* %6 to void ()*
  store atomic void ()* %7, void ()** @jlplt_ijl_set_task_tid_448994_got release, align 8
  musttail call void %6({} addrspace(10)* %0, i32 %1)
  ret void
}

loadfn=  %2 = load atomic void ()*, void ()** null unordered, align 8
opv=void ()** null
```

It seems to show up once I trigger parallel tests which is very strange. Before I run parallel tests, the same code just works...

wsmoses:
@avik-pal can you open an issue with the full dump?

wsmoses:
and can you see if using https://github.com/EnzymeAD/Enzyme.jl/pull/2068/files fixes it?

avik-pal:
Tests are now (mostly) happy. Need to look into downgrade eventually but I will call this a victory and merge

PR #1062, opened 2024-11-09
feat: nanoGPT implementation using Reactant

avik-pal:
### Needs

- [ ] `NNlib.gather` support in Reactant
- [ ] `NNlib.make_causal_mask`: https://github.com/EnzymeAD/Reactant.jl/pull/249


examples/NanoGPT/main.jl
```diff
@@ -0,0 +1,259 @@
...
+
+    alphabet, trainX, trainY, testX, testY = get_nanogpt_data(; sequence_length, test_split)
+
+    @printf "[Info] Alphabet size: %d\n" length(alphabet)
+    @printf "[Info] Training size: %d sequences.\n" size(trainX, 2)
+    @printf "[Info] Testing  size: %d sequences.\n\n" size(testX, 2)
+
+    train_loader = DataLoader(
+        (trainX, trainY); batchsize, shuffle=true, parallel=true
+    ) |> dev
```
Vaibhavdixit02, line 188:
```suggestion
    ) .|> dev
```

examples/NanoGPT/main.jl
```diff
@@ -0,0 +1,259 @@
...
+            time_taken = time() - start_time
+
+            if iter % 100 == 0
+                @printf "[Train] Epoch %3d\tIteration %6d\tLoss %.8e\tTime per \
+                         Iteration %0.5f\n" epoch iter loss time_taken
+            end
+        end
+
+        test_loss = loss_fn(
+            Array(first(model_compiled(testX, ps, Lux.testmode(st)))), testY
```
Vaibhavdixit02, line 232:
```suggestion
            first(model_compiled(testX, ps, Lux.testmode(st))), testY
```

examples/NanoGPT/main.jl
avik-pal: why is this needed? broadcasting over the dataloader will avoid constructing the DeviceIterator

examples/NanoGPT/main.jl
avik-pal: We need to compile the loss_fn for the non-`Array` version to work

examples/NanoGPT/main.jl
Vaibhavdixit02: Atleast, on GPU without this the labels weren't being transferred to the device

examples/NanoGPT/main.jl
Vaibhavdixit02: Ah yeah this might not be needed for your case here

PR #1111, opened 2024-11-30
Fix markdown list in docstring

abhro:
Trailing content is unindented, so the renderer reads the markdown list as fragmented. Indent the enclosed content by 2 spaces. Also dedent the lead sentences, so that there's less space to cover for trailing content. Thus, they don't get picked up as code by some other renderers.


abhro:
The checks seem to be failing because there isn't a name `backend` from GPUArrays.jl, which isn't what this PR addresses, so the CI failure should be unconnected AFAICT.

avik-pal (approved):

PR #1134, opened 2024-12-14
fix: use functors for testing wrapped arrays

avik-pal:


avik-pal:
Need to handle componentarrays before this can be merged

PR #1140, opened 2024-12-20
fix: update to new reactant changes

avik-pal:
Needs the following upstream changes:

- [x] New Tag in Reactant
- [x] https://github.com/EnzymeAD/Reactant.jl/pull/426 -- for eltype conversions
- [x] https://github.com/EnzymeAD/Reactant.jl/pull/417 -- for the new handling in the reactant extension
- [x] build reactant with 9.4 cudnn https://github.com/JuliaPackaging/Yggdrasil/pull/10176


avik-pal:
For #1143 we are applying a manual patch on the optimizers and we print a warning if we can't patch it yet

PR #1147, opened 2024-12-25
don't declare implicitly exported functions public

simeonschaub:
`@deprecate` by default exports the passed functions, <s>which I assume was not intended here</s>. This actually causes precompilation errors on 1.12 since these functions are also declared public, so remove the declaration as public

### DIFF
--- a/src/helpers/recursive_ops.jl
+++ b/src/helpers/recursive_ops.jl
@@ -110,3 +110,5 @@ For the following types it directly defines recursion rules:
     correctness of this implementation for specific usecases.
 """
 function recursive_map end
+
+@compat(public, (recursive_eltype,))
--- a/src/helpers/recursive_ops.jl
+++ b/src/helpers/recursive_ops.jl
@@ -110,7 +110,3 @@ For the following types it directly defines recursion rules:
     correctness of this implementation for specific usecases.
 """
 function recursive_map end
-
-@compat(public,
-    (recursive_add!!, recursive_copyto!, recursive_eltype,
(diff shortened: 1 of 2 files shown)


avik-pal:
These functions were deprecated in a `1.*` release, so we would have to keep the exports around I think. Why does it error in 1.12?

simeonschaub:
This error was added in https://github.com/JuliaLang/julia/pull/53664. If these exports were intentional I can also remove the public declaration in https://github.com/LuxDL/Lux.jl/blob/bcbcbce26b62f009ce22f567472f2ccee83af471/src/helpers/recursive_ops.jl#L114 since exported bindings are automatically public.

avik-pal:
> This error was added in [JuliaLang/julia#53664](https://github.com/JuliaLang/julia/pull/53664). If these exports were intentional I can also remove the public declaration in
> 
> https://github.com/LuxDL/Lux.jl/blob/bcbcbce26b62f009ce22f567472f2ccee83af471/src/helpers/recursive_ops.jl#L114
> 
> since exported bindings are automatically public.

lets remove the public here

src/helpers/recursive_ops.jl
```diff
@@ -110,7 +110,3 @@ For the following types it directly defines recursion rules:
     correctness of this implementation for specific usecases.
 """
 function recursive_map end
```
avik-pal, line 112:
```suggestion
function recursive_map end

@compat(public, (recursive_eltype,))
```

avik-pal (approved):

PR #1158, opened 2025-01-03
feat: migrate DDIM to Reactant

avik-pal:
Note we definitely need a new EnzymeJAX tag and a bunch of optimizations to go through before this can be run with the best performance


avik-pal:
This one definitely needs me to implement the loop using `@trace` else it takes forever to compile the generate function

avik-pal:
I have a feeling there is some Reactant gradient correctness issue here :cry:

avik-pal:
We don't encounter NaN gradients if we generate the noise from `rand_like` instead of `randn_like` but ofcourse that doesn't make too much sense

avik-pal:
@gemini-code-assist review

PR #1240, opened 2025-02-18
Update LUX.jl/examples/SimpleRNN/main.jl

aligurbu:
Don't I need to have "using Zygote" for this code to work since it is called `AutoZygote()` on line 149?

### DIFF
--- a/examples/SimpleRNN/main.jl
+++ b/examples/SimpleRNN/main.jl
@@ -9,9 +9,10 @@
 
 # ## Package Imports
 
-using ADTypes, Lux, JLD2, MLUtils, Optimisers, Printf, Reactant, Random
 # Note: If you wish to use AutoZygote() for automatic differentiation,
 # add Zygote to your project dependencies and include `using Zygote`.
+
+using ADTypes, Lux, JLD2, MLUtils, Optimisers, Printf, Reactant, Random
 
 # ## Dataset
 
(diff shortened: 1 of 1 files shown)


avik-pal:
> Don't I need to have "using Zygote" for this code to work since it is called AutoZygote() on line 149?

you are right. Let's add this as a comment, else we will have to add Zygote to the project deps, and we have been actively trying to cut down on the Zygote dependencies

aligurbu:
Hello Avik, 
I am unsure. Are the comments that I added what you meant to have instead? 
Thanks.

examples/SimpleRNN/main.jl
```diff
@@ -10,6 +10,8 @@
 # ## Package Imports
 
 using ADTypes, Lux, JLD2, MLUtils, Optimisers, Printf, Reactant, Random
+# Note: If you wish to use AutoZygote() for automatic differentiation,
+# add Zygote to your project dependencies and include `using Zygote`.
```
avik-pal, line 14:
```suggestion
# Note: If you wish to use AutoZygote() for automatic differentiation,
# add Zygote to your project dependencies and include `using Zygote`.

using ADTypes, Lux, JLD2, MLUtils, Optimisers, Printf, Reactant, Random
```

aligurbu (commented):
👌

avik-pal (approved):

PR #1242, opened 2025-02-18
feat: distinct recurrent initializer option for recurrent layers

### LINKED ISSUE #1241: Allow different initializers for input and hidden in recurrent layers
Currently, recurrent layers such as `RNNCell` and `LSTMCell` allow for just one initializer for both input and hidden weights. I want to use an Echo State Network where the hidden weights are initialized according to a custom function, but not the input weights; I don't train these layers.

I would like to specify initializer for input and weights separately. Is this something that could be needed also in the case of training? Let me know.

MartinuzziFrancesco:
Closes #1241. Also the default init for `GRUCell` is not `nothing` like specified in to docs but `glorot_uniform`, see [line 567](https://github.com/LuxDL/Lux.jl/blob/main/src/layers/recurrent.jl#L567). I don't think it makes a difference, but I just wanted to point it out since we are touching the initializers

### DIFF
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -185,7 +185,7 @@ applyrecurrentcell(l::AbstractRecurrentCell, x, ps, st, ::Nothing) = apply(l, x,
 
 @doc doc"""
     RNNCell(in_dims => out_dims, activation=tanh; use_bias=True(), train_state=False(),
-        init_bias=nothing, init_weight=nothing, init_recurrent_weight=nothing, init_state=zeros32)
+        init_bias=nothing, init_weight=nothing, init_recurrent_weight=init_weight, init_state=zeros32)
 
 An Elman RNNCell cell with `activation` (typically set to `tanh` or `relu`).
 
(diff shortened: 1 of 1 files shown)


src/layers/recurrent.jl
```diff
@@ -558,26 +574,30 @@ Gated Recurrent Unit (GRU) Cell
...
     init_weight
+    init_recurrent_weight
     init_state
     use_bias <: StaticBool
 end
 
 function GRUCell((in_dims, out_dims)::Pair{<:IntegerType, <:IntegerType};
         use_bias::BoolType=True(), train_state::BoolType=False(),
-        init_weight=glorot_uniform, init_bias=zeros32, init_state=zeros32)
+        init_weight=glorot_uniform, init_recurrent_weight=nothing, init_bias=zeros32, init_state=zeros32)
```
avik-pal, line 584: yes set it to glorot_uniform

avik-pal:
for the tests, you need to set the `init_recurrent_weight` as well. 

For this to be non_breaking, let's set `init_recurrent_weight=init_weight`

src/layers/recurrent.jl
MartinuzziFrancesco: ok, I'll update the old docstrings too

MartinuzziFrancesco:
you mean [here](https://github.com/LuxDL/Lux.jl/blob/main/test/layers/recurrent_tests.jl#L411)?

avik-pal:
> you mean [here](https://github.com/LuxDL/Lux.jl/blob/main/test/layers/recurrent_tests.jl#L411)? 

Yes. But if you default init recurrent weight to init weight this won't be an issue

avik-pal (approved):

PR #1261, opened 2025-03-12
Update OptimizationIntegration/main.jl comment to match link title

sinhtrung:
Changed Training Lux Models using Optimization.jl to Fitting with Optimization.jl to match the link name "Fitting with Optimization.jl" in the sidebar.


avik-pal (approved):
Thanks!

PR #1273, opened 2025-03-20
feat: support for ForwardDiff training

hstrey:
added LuxForwardDiffExt

It is working, but I may not have implemented the gradient and loss eval in the most efficient way


hstrey:
@avik-pal do you want me to add tests?

Project.toml
```diff
@@ -73,6 +73,7 @@ LuxMPIExt = "MPI"
 LuxMPINCCLExt = ["CUDA", "MPI", "NCCL"]
 LuxReactantExt = ["Enzyme", "Reactant"]
 LuxReverseDiffExt = ["FunctionWrappers", "ReverseDiff"]
+LuxForwardDiffExt = "ForwardDiff"
```
avik-pal, line 76: ForwardDiff is already a main dependency, you can add the training dispatch to the main codebase without the extensions

hstrey: @avik-pal Should this training dispatch go into src/helpers/training.jl?

src/helpers/training.jl
```diff
@@ -160,6 +161,7 @@ Compute the gradients of the objective function wrt parameters stored in `ts`.
 | `AutoReverseDiff(; compile)` | `ReverseDiff.jl` |
 | `AutoTracker`                | `Tracker.jl`     |
 | `AutoEnzyme`                 | `Enzyme.jl`      |
+| `AutoForwardDiff`            | `ForwardDiff.jl` |
```
avik-pal, line 164:
```suggestion
| `AutoForwardDiff`            |                  |
```

we already have it loaded so no extra packages are needed

src/helpers/training.jl
```diff
@@ -1,13 +1,14 @@
...
-using ADTypes: AbstractADType, AutoEnzyme, AutoReverseDiff, AutoTracker, AutoZygote
+using ADTypes: AbstractADType, AutoEnzyme, AutoReverseDiff, AutoTracker, AutoZygote, AutoForwardDiff
 using Compat: @compat
 using ConcreteStructs: @concrete
 using FastClosures: @closure
 using Functors: Functors, fmap
 using Optimisers: Optimisers
 using Setfield: @set!
 using Static: StaticBool, Static, False, True, static
+using ForwardDiff
```
avik-pal, line 11:
```suggestion
using ForwardDiff: ForwardDiff
```

avik-pal (approved):
Overall looks good, lets add a test for this

hstrey:
@avik-pal I added AutoForwardDiff to the training test.  I was not sure whether ForwardDiff works on gpus, so I added AutoForwardDiff to the list of adtypes that are not run on gpus - at least this is how I understood the code.

hstrey:
@avik-pal I checked the logs.  The ForwardDiff fails because ForwardDiff.gradient chokes on that fact that ts.parameters is a named tuple.  When I tested it, I had used ComponentVector.  Should I do that here too?

hstrey:
@avik-pal, all the tests I added involved named tuples. Should I write a separate test to test just the ForwardDiff training interface?

avik-pal:
> @avik-pal, all the tests I added involved named tuples. Should I write a separate test to test just the ForwardDiff training interface?

yeah

test/helpers/training_tests.jl
```diff
@@ -139,6 +139,58 @@ end
...
+
+    for epoch in 1:1000, (x, y) in dataset_
+        grads, loss, _, tstate = allow_unstable() do
+            Training.single_train_step(AutoForwardDiff(), mse, (x, y), tstate)
+        end
+    end
+
+    final_loss = first(
+        mse(model, tstate.parameters, tstate.states, dataset_[1])
+    )
```
avik-pal, line 191: add a check for final_loss < ...

src/helpers/training.jl
```diff
@@ -208,6 +212,15 @@ function maybe_wrap_adtype(
                          Enzyme.jl (`AutoEnzyme`)."))
 end
 
+function compute_gradients_impl(
+    ::AutoForwardDiff, objective_function::F, data, ts::Lux.Training.TrainState
+) where {F}
+    grads = ForwardDiff.gradient(ps -> objective_function(ts.model,ps,ts.states,data)[1], ts.parameters);
+    loss, st, stats = objective_function(ts.model,ts.parameters,ts.states,data)
+    @set! ts.states = st
```
avik-pal, line 220: This is incorrect, we shouldn't do 2 evals of `objective_function` here (apart from being inefficient, it will actually give incorrect results for normalization layers). The easiest way would be to use a closure approach similar to how it is done in the enzyme extension

hstrey: @avik-pal I did what you suggested, but I am not completely sure whether my implementation is sound.  I used the DiffResults interface for ForwardDiff to get both the value and the grad at the same time.  I also used the wrapper that is used by the Enzyme extension. 
On another note, when I benchmarked the different AD interfaces, for small NNs ForwardDiff can be significantly faster.  I found that for our use case, we are 4x faster.  Maybe it makes sense to implement ForwardDiff in a way that takes care of the named tuple issue by applying ComponentVector when necessary.  But that would mean that ComponentVector becomes a dependency.  One could also fix it at the ForwardDiff level.
The tests should pass now.

src/helpers/training.jl
```diff
@@ -160,6 +161,8 @@ Compute the gradients of the objective function wrt parameters stored in `ts`.
 | `AutoReverseDiff(; compile)` | `ReverseDiff.jl` |
 | `AutoTracker`                | `Tracker.jl`     |
 | `AutoEnzyme`                 | `Enzyme.jl`      |
+| `AutoForwardDiff`            |                  |
+
```
avik-pal, line 165:
```suggestion
```

src/helpers/training.jl
```diff
@@ -8,6 +8,7 @@ using Functors: Functors, fmap
 using Optimisers: Optimisers
 using Setfield: @set!
 using Static: StaticBool, Static, False, True, static
+using ForwardDiff: ForwardDiff, DiffResults
```
avik-pal, line 11:
```suggestion
```

src/helpers/training.jl
```diff
@@ -185,6 +188,7 @@ A 4-Tuple containing:
   - `AutoReverseDiff(; compile=true)` is not supported for Lux models with non-empty state
     `st`. Additionally the returned stats must be empty (`NamedTuple()`). We catch these
     issues in most cases and throw an error.
+  - AutoForwardDiff only works with parameters that are converted to ComponentVectors (e.g. ps=ComponentVector(ps))
```
avik-pal, line 191:
```suggestion
  - AutoForwardDiff only works with parameters that are AbstractArrays
    (e.g. ps=ComponentVector(ps))
```

PR #1303, opened 2025-04-16
fix: new reactant version

avik-pal:


avik-pal:
waiting on https://github.com/JuliaPackaging/Yggdrasil/pull/11016 & corresponding reactant tag

PR #1318, opened 2025-05-03
allow SelectDim to take arbitrary views

ExpandingMan:
This allows `SelectDim` to take any view.  This functionality is already supported by `Base.selectdim` so it only requires loosening type restrictions.  I allowed it to use arbitrary `AbstractVector`s but obviously using completely arbitrary non-contiguous views is inadvisable in most cases (it's still not entirely clear to me how good an idea it is to use `view` in deep learning on GPU's anyway).

### DIFF
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -202,7 +202,7 @@ SelectDim(dim::Integer, index::AbstractVector) = SelectDim(static(dim), index)
 function (s::SelectDim{D,<:StaticInt})(x, _, st::NamedTuple) where {D}
     return selectdim(x, known(s.dim), known(s.index)), st
 end
-(s::SelectDim)(x, _, st::NamedTuple) = selectdim(x, known(s.dim), s.index)
+(s::SelectDim)(x, _, st::NamedTuple) = selectdim(x, known(s.dim), s.index), st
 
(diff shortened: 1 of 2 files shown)


src/layers/basic.jl
```diff
@@ -191,12 +193,16 @@ Return a view of all the data of the input `x` where the index for dimension `di
...
 
-SelectDim(dim, index) = SelectDim(static(dim), static(index))
+SelectDim(dim::Integer, index::Integer) = SelectDim(static(dim), static(index))
+SelectDim(dim::Integer, index::AbstractVector) = SelectDim(static(dim), index)
 
-(s::SelectDim)(x, _, st::NamedTuple) = selectdim(x, known(s.dim), known(s.index)), st
+function (s::SelectDim{D,<:StaticInt})(x, _, st::NamedTuple) where {D}
+    return selectdim(x, known(s.dim), known(s.index)), st
+end
+(s::SelectDim)(x, _, st::NamedTuple) = selectdim(x, known(s.dim), s.index)
```
avik-pal, line 205:
```suggestion
(s::SelectDim)(x, _, st::NamedTuple) = selectdim(x, known(s.dim), s.index), st
```

avik-pal (approved):

PR #1385, opened 2025-07-09
feat: annotate important parts of training loop

avik-pal:
- [x] Locally see the tracing
- [x] Add to documentation

### DIFF
--- a/ext/LuxReactantExt/training.jl
+++ b/ext/LuxReactantExt/training.jl
@@ -16,10 +16,9 @@ function compute_gradients_internal(objective_function::F, model, data, ps, st) 
     return dps, loss, stats, stₙ
 end
 
-function Lux.Training.compute_gradients_impl(
+Profiler.@annotate "Compile Compute Gradients" function Lux.Training.compute_gradients_impl(
     backend::ReactantBackend, objective_function::F, data, ts::Training.TrainState
 ) where {F}
-    compile_start_time = time()
     compiled_gradient_function = Reactant.with_config(;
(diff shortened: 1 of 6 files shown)


avik-pal:
```julia
using Reactant, Lux, Random, MLUtils, Optimisers

dev = reactant_device()

x_data = rand(Float32, 32, 1024)
y_data = x_data .^ 2 .- 1

dl = DataLoader((x_data, y_data); batchsize=32, shuffle=true) |> dev;

model = Chain(Dense(32 => 64, relu), Dense(64 => 32, relu), Dense(32 => 1))
ps, st = Lux.setup(Random.default_rng(), model) |> dev;

Reactant.with_profiler("/tmp/traces/lux_training/") do
    train_state = Training.TrainState(model, ps, st, Adam(0.001))
    for epoch in 1:10
        for (x, y) in dl
            _, loss, _, train_state = Training.single_train_step!(
                AutoEnzyme(), MSELoss(), (x, y), train_state; return_gradients=Val(false)
            )
            @show loss
        end
    end
end
```

ext/LuxReactantExt/training.jl
```diff
@@ -16,10 +16,9 @@ function compute_gradients_internal(objective_function::F, model, data, ps, st)
     return dps, loss, stats, stₙ
 end
 
-function Lux.Training.compute_gradients_impl(
+Profiler.@annotate "Compile Compute Gradients" function Lux.Training.compute_gradients_impl(
```
wsmoses, line 19: cc @jumerckx we should likely consider adding this to all function calls [and also ops.jl]

PR #1471, opened 2025-09-09
feat: don't collect FillArrays for reactant

### LINKED ISSUE #1470: MLDataDevices: Keep FillArray uncollected for some device types?
Currently, MLDataDevices always collects an `AbstractFill` if not on CPU:

```
Adapt.adapt_structure(::CPUDevice, x::AbstractFill) = x
Adapt.adapt_structure(to::AbstractDevice, x::AbstractFill) = Adapt.adapt(to, collect(x))
```

Can we make exceptions, depending on the device type? Reactant, for example, can handle `FillArray`s now (https://github.com/EnzymeAD/Reactant.jl/pull/1568).

CC @avik-pal

avik-pal:
fixes #1470

### DIFF
--- a/lib/MLDataDevices/Project.toml
+++ b/lib/MLDataDevices/Project.toml
@@ -1,7 +1,7 @@
 name = "MLDataDevices"
 uuid = "7e8f7934-dd98-4c1a-8fe8-92b47a384d40"
 authors = ["Avik Pal <avikpal@mit.edu> and contributors"]
-version = "1.11.3"
+version = "1.12.0"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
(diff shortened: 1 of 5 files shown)


oschulz:
Wow, that was quick! And a lot less trivial than I though it would be, from the looks of it - thanks @avik-pal !

avik-pal:
> Wow, that was quick! And a lot less trivial than I though it would be, from the looks of it - thanks @avik-pal !

It would have been easier if all of the types carried around some form of data, but Zeros/Ones only indicate device based on type. Had to do the full infra for handling those cases

oschulz:
What about CUDA in regard to FillArrays? Can it handle it reliably?

avik-pal:
I don't think CUDA (not reactant's cuda backend) works well with FillArrays (though I haven't tested it in sometime)

oschulz:
> I don't think CUDA (not reactant's cuda backend) works well with FillArrays (though I haven't tested it in sometime)

Still seems to be the case. I played around a bit, broadcasts with a mix of `CuArray` and `FillArray` arguments work surprisingly well, but even a simple matmul fails, for example:

```julia
julia> CuArray(rand(Float32, 100, 100)) * Fill(Float32(1), 100, 100)
ERROR: Scalar indexing is disallowed.
```

PR #1485, opened 2025-09-18
feat: precompile common workloads

avik-pal:
- [x] precompile workloads
- [x] re-enable all testing
- [x] remove debug logging


avik-pal:
Reactant is crashing for reasons that I don't fully understand

PR #1542, opened 2025-11-03
refactor: use `EnzymeRules.@easy_rule` in Lux.jl

avik-pal:


lib/LuxLib/src/impl/matmul.jl
```diff
@@ -244,105 +244,15 @@ function CRC.rrule(
...
-
-    if !(C isa EnzymeCore.DuplicatedNoNeed || C isa EnzymeCore.BatchDuplicatedNoNeed)
-        matmuladd!(C.val, opmode.val, A.val, B.val, bias.val)
-    end
-
-    return EnzymeRules.AugmentedReturn(nothing, nothing, (A_cache, B_cache))
-end
 
-function EnzymeRules.reverse(
-    cfg,
```
wsmoses, line 285: the in place manual rules are still going to be more efficient than the easy rule

yebai:
Would it make sense to write these rules using `ChainRules`, then import them from AD backends (e.g., Enzyme, Mooncake)? 

See https://github.com/chalk-lab/Mooncake.jl/issues/842#issuecomment-3508208031 and @ChrisRackauckas's follow-up comments.

avik-pal:
> Would it make sense to write these rules using ChainRules, then import them from AD backends (e.g., Enzyme, Mooncake)?

The chain rules for these already exist

lib/LuxLib/src/impl/matmul.jl
avik-pal: It probably makes sense to keep these then.  I will get rid of the oop versions that had these custom rules with easy_rule

PR #1545, opened 2025-11-09
test: use finite differencing for gradient testing for Reactant

avik-pal:
needs https://github.com/EnzymeAD/Reactant.jl/pull/1846 + a version bump.

This is also one of the important bits before we can start testing on 1.12


lib/LuxLib/test/Project.toml
```diff
@@ -50,7 +50,7 @@ ComponentArrays = "0.15.22"
 Enzyme = "0.13.81"
 EnzymeCore = "0.8.14"
 ExplicitImports = "1.9.0"
-ForwardDiff = "0.10.36, =1"
+ForwardDiff = "0.10.36, 1"
```
avik-pal:
```suggestion
ForwardDiff = "0.10.36, =1"
```

PR #1605, opened 2025-12-27
fix: forwarddiff support for gpu arrays

### LINKED ISSUE #1530: Relax ForwardDiff version bound?
Lux is currently very strict with ForwardDiff:

```
ForwardDiff = "0.10.36, =1"
```

which keeps ForwardDiff back a lot in some package combination. Bijectors, for example, has

```
ForwardDiff = "0.10, 1.0.1"
```

so when installing both Lux and Bijectors, one currently ends up with ForwardDiff v0.10.36.

avik-pal:
fixes https://github.com/LuxDL/Lux.jl/issues/1530


oschulz:
Thanks a lot @avik-pal !

PR #1645, opened 2026-01-28
Fixed legend labels for the curves in the PolynomialFitting documentation page

JamieMair:
Current visualisation [here](https://lux.csail.mit.edu/stable/tutorials/beginner/2_PolynomialFitting#Dataset) has the wrong labels (they should be switched).

PR simply switches the labels

### DIFF
--- a/examples/PolynomialFitting/main.jl
+++ b/examples/PolynomialFitting/main.jl
@@ -48,7 +48,7 @@ begin
         color=:orange,
         strokecolor=:black,
         strokewidth=2,
-        label="Data Points",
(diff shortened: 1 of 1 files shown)


examples/PolynomialFitting/main.jl
```diff
@@ -48,7 +48,7 @@ begin
         color=:orange,
         strokecolor=:black,
         strokewidth=2,
-        label="True Quadratic Function",
+        label="Data Points",
```
avik-pal:
```suggestion
        label="Actual Data",
```

PR #1711, opened 2026-05-05
fix: CUDAExt for WeightInitializers

### LINKED ISSUE #1709: WeightInitializers CUDAExt errors
When trying to use Lux with CUDA/cuDNN (to train with Zygote GPU), `WeightInitializers` fails to precompile: 

```julia
julia> using WeightInitializers
[ Info: Precompiling CUDAExt [0d0e9cde-cfc5-56bb-a277-295c1cc0b950](cache misses: wrong dep version loaded (2), incompatible header (2))
Info Given CUDAExt was explicitly requested, output will be shown live 
ERROR: LoadError: UndefVarError: `RNG` not defined in `cuRAND`
Suggestion: check for spelling errors or missing imports.
Hint: a global variable of this name also exists in CUDA.
```

This happens with with latest CUDA releases (v6); perhaps some incompatibility with the CUDA rework into sub-packages?

Pkg status:
```julia
Status `~/evovest/NeuroTabModels.jl/adhoc/Project.toml`
  [052768ef] CUDA v6.1.0
  [b2108857] Lux v1.31.4
  [d49dbf32] WeightInitializers v1.3.3
  [02a925ec] cuDNN v6.1.0
```

avik-pal:
fixes https://github.com/LuxDL/Lux.jl/issues/1709

### DIFF
--- a/lib/WeightInitializers/Project.toml
+++ b/lib/WeightInitializers/Project.toml
@@ -1,7 +1,7 @@
 name = "WeightInitializers"
 uuid = "d49dbf32-c5c2-4618-8acc-27bb2598ef2d"
 authors = ["Avik Pal <avikpal@mit.edu> and contributors"]
-version = "1.3.3"
+version = "1.3.4"
 
 [deps]
 ConcreteStructs = "2569d6c7-a4a2-43d3-a901-331e8e4be471"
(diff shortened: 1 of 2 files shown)


nicholaskl97:
Hello, I see that this version of WeightInitializers (v1.3.4) was merged into the main branch, but it appears that hasn't been [registered](https://github.com/JuliaRegistries/General/blob/master/W/WeightInitializers/Versions.toml) yet. Is that intentional? I came across the precompilation failure this is supposed to fix today and it looks like the fix is ready and merged, but not yet released.

PR #1739, opened 2026-07-12
Try to re-enable enzyme tests on 1.11+

wsmoses:


ChrisRackauckas-Claude:
For what it is worth, none of the red on this PR looks like it comes from re-enabling the Enzyme tests — every actual test group passed, including `test (1.12, others)`, `autodiff`, `core_layers`, `recurrent_layers` and `reactant` on both 1.10 and 1.12. The twelve failures split into two pre-existing problems:

**1. The six `downgrade` jobs — stale compat floors, unrelated to this diff.** Your run has been failing since 2026-07-12 with:

```
ERROR: LoadError: LLVM error: unknown function pass 'custom-nvvm-reflect'
Failed to precompile ReactantCUDAExt [91356394-6239-50cb-9e6b-6f1f3f798cde]
```

Reactant below 0.2.256 asks LLVM.jl for a pass name that current LLVM.jl no longer registers, and Lux floors Reactant at 0.2.205 (test) / 0.2.236 (package). I audited every declared floor against every other floor's requirements (registry `Compat.toml` **and** `WeakCompat.toml`, both directions) and found twelve stale pins, including three that cap `ForwardDiff` at 0.10 while Lux floors it at 1, and two that cap `AMDGPU` below the declared 2.2 floor. Fix is up at https://github.com/LuxDL/Lux.jl/pull/1749 — compat-only, no code. If that merges, these six should go green here on a rebase.

**2. `test (1.10, misc)` and `test (1.12, misc)` — Aqua, also unrelated.**

```
4 ambiguities found. To get a list, set `broken = false`.
Test Failed at ~/.julia/packages/Aqua/h1qD0/src/ambiguities.jl:80
```

The ambiguities are between `Lux.NilSizePropagation.Nil`'s `convert` and `CommonWorldInvalidations.Despec4`, a transitive dependency that deliberately defines despecialization types. Needs an exclusion or a disambiguating method; not a floor issue and not yours. I have not fixed that one — happy to if it is wanted.

One thing you may hit after the CI rot clears, since it is adjacent to what this PR is testing: on Julia 1.12 the newest Enzyme (0.13.199) fails to precompile inside `LuxTestUtils` with

```
UndefVarError: `runtime_slug` not defined in `GPUCompiler`
  @ ~/.julia/packages/Enzyme/OvStQ/src/compiler.jl:181
```

Enzyme 0.13.199 declares `GPUCompiler = "1.6.2"`, so this shows up wherever a newer GPUCompiler gets in. That is what currently sinks `test (1.12, others)` on other PRs, and no floor bump can fix it from the Lux side.

Context for why I was in here: SimpleChains 0.5.0 ships an Enzyme reverse rule (PumasAI/SimpleChains.jl#220), and https://github.com/LuxDL/Lux.jl/pull/1748 widens Lux's SimpleChains compat to admit it.

PR #1749, opened 2026-08-26
fix(ci): raise Reactant and CUDA floors so downgrade tests resolve

ChrisRackauckas-Claude:
## What changed and why

All six `downgrade` jobs currently fail, on every PR, while precompiling `ReactantCUDAExt`:

```
ERROR: LoadError: LLVM error: unknown function pass 'custom-nvvm-reflect'
  [3] macro expansion
    @ ~/.julia/packages/Reactant/.../ext/ReactantCUDAExt.jl:691 [inlined]
  [5] vendored_optimize_module!(job::Any, mod::LLVM.Module, instcombine::Bool)
```

Reactant below 0.2.256 asks LLVM.jl for a pass name that the LLVM.jl it resolves against today (9.13.0) no longer registers. `test/Project.toml` floors Reactant at `0.2.205` and the package floors it at `0.2.236`, so under forced-lowest resolution the extension cannot precompile. `0.2.257` is the first version that does.

Separately, the declared `CUDA` floor was unsatisfiable against the `Reactant` floor — Reactant 0.2.191+ requires CUDA ≥ 5.9, so forcing both to their lower bounds leaves CUDA with nothing:

```
ERROR: Unsatisfiable requirements detected for package CUDA [052768ef]:
 ├─restricted to versions 5.8.0 by an explicit requirement, leaving only versions: 5.8.0
 └─restricted by compatibility requirements with Reactant [3c362404] to versions: 5.9.0 - 5.11.3 — no versions left
```

Both are lower-bound corrections only; no code changes.

## Verification

Bisected the Reactant boundary on Julia 1.11 by pinning Reactant and CUDA and precompiling the extension. `FAIL` is the `custom-nvvm-reflect` abort above:

| Reactant | CUDA | Result |
|---|---|---|
| 0.2.205 (test floor) | 5.9.6 | FAIL |
| 0.2.236 (package floor) | 5.9.0 | FAIL |
| 0.2.236 | 5.11.3 | FAIL |
| 0.2.240 / 0.2.245 / 0.2.253 / 0.2.255 | 5.9–5.11.3 | FAIL |
| 0.2.256 | 5.11.3 | fails differently: `UndefVarError: fpm not defined in ReactantCUDAExt` |
| **0.2.257** | **5.9.6** | **OK** (LLVM 9.13.0) |
| 0.2.257 / 0.2.259 / 0.2.260 | 5.11.3 | OK |

CUDA is left at `5.9.6` in `test/Project.toml` (already above the new package floor) and raised to `5.9` in `Project.toml`.

## What I did not verify

- I did not run a full `downgrade` job end-to-end locally. I ran this repo's own `julia-actions/julia-downgrade-compat` `downgrade.jl` in `forcedeps` mode on Julia 1.11, which reported `All forcedeps checks passed`, but the subsequent `Pkg.test` re-resolves and pulls Reactant 0.2.283 rather than the floor, and it then failed in my sandbox for an unrelated reason (CondaPkg/pixi could not reach the network). The evidence above is therefore the pinned-version matrix, which reproduces the CI error exactly and shows the first version that clears it.
- GPU paths, and whether some Reactant version between 0.2.257 and current would be a better floor for other reasons.

## Not fixed here, but also red on `main` today

Flagging these so they are not mistaken for fallout from this PR — none is a bounds problem:

1. `test (1.10, misc)` / `test (1.12, misc)` — Aqua reports 4 ambiguities between `Lux.NilSizePropagation.Nil`'s `convert` and `CommonWorldInvalidations.Despec4`, a transitive dependency that deliberately defines despecialization types. Needs an exclusion or a disambiguating method, not a floor.
2. `test (1.12, others)` — the Flux downstream test dies because Enzyme **0.13.199**, the newest release, calls `GPUCompiler.runtime_slug` (`Enzyme/src/compiler.jl:181`), which no longer exists. There is no newer Enzyme to raise a floor to; this needs an upper bound or an upstream fix.
3. `airspeed-velocity` and the two `buildkite` checks fail on fork-PR permissions (403 / pipeline upload) and should pass on a branch in this repo.

I am happy to follow up on 1 and 2 in separate PRs if that is useful.

This PR was authored by an agent on behalf of @ChrisRackauckas.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Q4CWNHT9we9k1tFL4hSXrM


ChrisRackauckas-Claude:
Pushed `f324ba96`. The first commit was not enough — CI came back with a different failure, so I audited the floors properly instead of fixing them one CI round at a time.

**What the first attempt missed.** Raising `Reactant` to 0.2.257 moved the failure rather than removing it:

```
ERROR: LoadError: Unsatisfiable requirements detected for package ReactantCore [a3311ec8]:
 ├─restricted to versions 0.1.17 by an explicit requirement, leaving only versions: 0.1.17
 └─restricted by compatibility requirements with Reactant [3c362404] to versions: 0.1.18 - 0.1.21 — no versions left
```

I had verified the new Reactant floor in an isolated environment where every *other* package was free to move, so nothing pinned `ReactantCore` down. That was the wrong test.

**The audit.** I wrote a checker that takes every floor declared in `Project.toml` and `test/Project.toml` (intersected, higher wins), then for each one reads that exact version's registry `Compat.toml` **and `WeakCompat.toml`** and asks whether every other declared floor satisfies it — in both directions, so a floor that is too *high* for another package's upper bound is caught as well. Ten floors were stale:

| Package | Was | Now | Why |
|---|---|---|---|
| `ReactantCore` | 0.1.17 | 0.1.18 | Reactant 0.2.257 requires it |
| `StableRNGs` | 1.0.2 | 1.0.4 | Reactant 0.2.257 requires it |
| `OneHotArrays` | 0.2.5 | 0.2.10 | Reactant 0.2.257 requires it |
| `PythonCall` | 0.9 | 0.9.25 | Reactant 0.2.257 requires it |
| `GPUArrays` | 11 | 11.3.1 | CUDA 5.9.6 requires it |
| `Zygote` | 0.7 | 0.7.5 | 0.7.0 caps ForwardDiff at 0.10, but ForwardDiff floors at 1 |
| `LoopVectorization` | 0.12.171 | 0.12.172 | same ForwardDiff cap |
| `SimpleChains` | 0.4.7 | 0.4.8 | same ForwardDiff cap |
| `Flux` | 0.16.3 | 0.16.5 | 0.16.3 caps AMDGPU at 1, but AMDGPU floors at 2.2 |
| `MPI` | 0.20.19 | 0.20.23 | 0.20.19 caps AMDGPU at 0.8 |

Plus the two from the first commit: `Reactant` 0.2.236/0.2.205 → 0.2.257 (below 0.2.256 `ReactantCUDAExt` asks LLVM.jl for the removed `custom-nvvm-reflect` pass; 0.2.256 itself is broken with `UndefVarError: fpm`), and `CUDA` 5.8 → 5.9 (Reactant 0.2.191+ requires ≥ 5.9).

**Verification.** The audit reports `NO FLOOR CONFLICTS`, and — this time with the local `lib/*` packages `develop`ed, the way CI builds the environment rather than in isolation — the full pinned set resolves and precompiles on Julia 1.11:

```
CUDA 5.9.6   Flux 0.16.5      ForwardDiff 1.0.0   GPUArrays 11.5.13  LLVM 9.13.1
OneHotArrays 0.2.10  Reactant 0.2.257  ReactantCore 0.1.18  SimpleChains 0.4.8  Zygote 0.7.5
```

No `custom-nvvm-reflect` abort, no unsatisfiable requirement.

**Still not verified**: a full downstream `downgrade` job end-to-end locally — `Pkg.test` in my sandbox dies in CondaPkg/pixi for reasons unrelated to this change. CI on this PR is the real check, and if it finds another stale floor the audit script will now name it directly rather than costing another round trip.

**Note for the `lib/*` packages**: they carry the same stale floors (`lib/LuxLib` `Reactant = "0.2.170"`, `ReactantCore = "0.1.15"`, `lib/MLDataDevices` and `lib/LuxCore` `Reactant = "0.2.203"`, `lib/*/CUDA = "5.8, 6"`, `lib/LuxCore` `Flux = "0.16.3"`). They do not break Lux's own downgrade job, because the merged floor is the highest of the declarations, but their own downgrade jobs will hit this the moment those packages are touched. Happy to send that as a separate PR.

avik-pal (approved):

PR #1754, opened 2026-08-29
feat: use updated cuDNN bindings

avik-pal:


lib/LuxLib/Project.toml
```diff
@@ -73,7 +73,7 @@ AppleAccelerate = "0.4, 0.5"
 ArrayInterface = "7.17.1"
 BLISBLAS = "0.1, 0.2"
 CPUSummary = "0.2.6"
-CUDA = "5.8, 6"
+CUDA = "6.3"
```
wsmoses, line 76: @avik-pal note this is going to cause problems for merging at the moment, as CUDA 6.3 forces GPUCompiler2, which is not yet supported by Enzyme or reactant

